PR #10, opened 2023-10-11
Just demoing git functiunality

paulseidel08:
demo only. no need to review


atrejut:
please change the PR title -- it's an exercise submission not just the demo PR

PR #17, opened 2023-10-11
Übung1

not-too-deranged:


Exercises/JasminSteidl/JasminSteidlÜ1.py
```diff
@@ -0,0 +1,43 @@
+# use this template for completing todays exercises
+# make a copy of the file, in a folder with your own name
+# add code below to answer the questions; for each question there should be
+# single print() which will give the answer(s) to that question.
+
+# what are the quotient and remainder after (integer) division of 2711274328912 by 23369245575
+x=2711274328912
+y=23369245575
+print(x/y)
```
christianrietzsch, line 9: use `x // y` for integer division instead of `x / y` for floating point division

Exercises/JasminSteidl/JasminSteidlÜ1.py
```diff
@@ -0,0 +1,43 @@
...
+# which and how many elements are common to the two lists above? 
+a = set(first_list)
+b = set(second_list)
+if a==b: print(first_list) 
+else: print()
+print(len(first_list))
+
+
+# how often does the character 'a' occur in each of the lists above?
+lista = str ("".join(first_list))
```
christianrietzsch, line 40: the type casting from `str` to `str` is unnecessary in `str ("".join(first_list))` and `str ("".join(second_list))`
also listb should contain the string value from the second list `listb = "".join(second_list)` and not `listb = "".join(first_list)`

Exercises/JasminSteidl/JasminSteidlÜ1.py
```diff
@@ -0,0 +1,43 @@
...
+# how may elements are in each of the following lists?
+first_list = ["ybk", "zcg", "dhw", "gua", "mqh", "yel", "fpk", "imk", "nkt", "tpl", "guc", "ozk", "dip", "thn", "pxc", "obk", "orl", "rdc", "vto", "wen", "tbw", "itj", "qfb", "ymd", "txb", "vgw", "wgd", "tcg", "ezm", "gnt", "ehf", "pfo", "alf"]
+second_list = ['ybk', 'zcg', 'dhw', 'gua', 'mqh', 'yel', 'fpk', 'imk', 'nkt', 'tpl', 'guc', 'ozk', 'dip', 'thn', 'pxc', 'obk', 'orl', 'rdc', 'vto', 'wen', 'tbw', 'itj', 'qfb', 'ymd', 'txb', 'vgw', 'wgd', 'tcg', 'ezm', 'gnt', 'ehf', 'pfo', 'alf']
+print(len(first_list))
+print(len(second_list))
+
+# which and how many elements are common to the two lists above? 
+a = set(first_list)
+b = set(second_list)
+if a==b: print(first_list)
```
christianrietzsch, line 34: `if a==b` only checks if the lists are completely equal but the task was to find out how many / which elements are common: 
`common_list = a.intersection(b)`
`print(f"there are {len(common_list)} common elements: {common_list}")`

christianrietzsch (changes requested):

PR #22, opened 2023-10-11
Exercise 01 / Utku Acikel

UtkuDHBW:
Exercise 01 Finished


atrejut (changes requested):
You are modifying `Vorlesungsunterlagen/04_Freitag.ipynb` in this PR, as well as adding a bunch of `.DS_Store` files which should not be commited...

PR #24, opened 2023-10-12
Lösungen 01

s231365:
Bitte reviewen


fabischw (approved):
Looks good, consider printing more information in line 36 and 37 to let the user know what the data is showing (only the set and the length of the set are being printed and the user / reviewer has to go into the code to find out what is being printed) next time. You might also want to look into using f-strings in python instead of using commas in the print statements in the future.
Great solution overall.

PR #36, opened 2023-10-14
Solution to Task 2 Julius Dralle

RubelOS:


Exercises/JuliusDralle/exercises_02.py
```diff
@@ -0,0 +1,88 @@
...
+# Think of some good inputs to test this functionality, write down at least three
+# examples and verify that the output for these examples is correct.
+
+def splittingMachine(*args):
+    numberList = []
+    charList = []
+    for i in args:
+        if i >= "0" and i <= "9":
+            numberList.append(i)
+        else:
```
thecodingcube77, line 80: You don't cover all cases. What if the given argument is no number but also not a single character?

Exercises/JuliusDralle/exercises_02.py
```diff
@@ -0,0 +1,88 @@
...
+            return -1
+    newList.append(returnValue)
+    newList.append(args[1])
+    newList.append(args[2])
+    newList.append(args[3])
+    return 0
+    
+
+
+def calculatorMachine( memoryList, *args):
```
thecodingcube77, line 52: Your whole calculatorMachine() is not working as required in the task, because you never mutate the memoryList passed into the function. This is needed for you to get the right solution.
This idea is also quite overcomplicated, you should not really need a second function.

Exercises/JuliusDralle/exercises_02.py
```diff
@@ -0,0 +1,88 @@
...
+    numberList = []
+    charList = []
+    for i in args:
+        if i >= "0" and i <= "9":
+            numberList.append(i)
+        else:
+            charList.append(i)
+    return numberList, charList
+
+print(splittingMachine(*["1","a","b","5","6"]))
```
thecodingcube77, line 84: You dont't cover all cases in your tests. You should pass a string consisting out of numbers and letters into the splittingMachine() function.
E.g.: "34sdfKHkls" which should then be completely filtered out.

thecodingcube77 (changes requested):
Please make it work like required as stated in ma comments. You can look at my solution or ask me if you have further questions.

thecodingcube77 (approved):
Now it's all correct

DrKeks1010 (approved):
Well done

Exercises/JuliusDralle/exercises_02.py
```diff
@@ -0,0 +1,73 @@
...
+# and ending the program. The final memory looks like [2, 0, 0, 0, 99]. Your function should return 2.
+
+# Here's another testcase:
+# [1, 1, 1, 4, 99, 5, 6, 0, 99] should become [30, 1, 1, 4, 2, 5, 6, 0, 99]
+# Your function should return 30.
+
+# print out which value is returned by your function for the following list:
+commands = [1, 12, 2, 3, 1, 1, 2, 3, 1, 3, 4, 3, 1, 5, 0, 3, 2, 1, 9, 19, 1, 5, 19, 23, 1, 6, 23, 27, 1, 27, 10, 31, 1, 31, 5, 35, 2, 10, 35, 39, 1, 9, 39, 43, 1, 43, 5, 47, 1, 47, 6, 51, 2, 51, 6, 55, 1, 13, 55, 59, 2, 6, 59, 63, 1, 63, 5, 67, 2, 10, 67, 71, 1, 9, 71, 75, 1, 75, 13, 79, 1, 10, 79, 83, 2, 83, 13, 87, 1, 87, 6, 91, 1, 5, 91, 95, 2, 95, 9, 99, 1, 5, 99, 103, 1, 103, 6, 107, 2, 107, 13, 111, 1, 111, 10, 115, 2, 10, 115, 119, 1, 9, 119, 123, 1, 123, 9, 127, 1, 13, 127, 131, 2, 10, 131, 135, 1, 135, 5, 139, 1, 2, 139, 143, 1, 143, 5, 0, 99, 2, 0, 14, 0]
+
+def calculate():
```
fabischw, line 31: Your function for exercise one does not take an array as an input as stated in the instructions and instead uses the commands defined above. Please change your function to take an input as stated in the requirements.

Exercises/JuliusDralle/exercises_02.py
```diff
@@ -0,0 +1,73 @@
...
+# Think of some good inputs to test this functionality, write down at least three
+# examples and verify that the output for these examples is correct.
+
+def splittingMachine(*args):
+    numberList = []
+    charList = []
+    for i in args:
+        if i.isdigit():
+            numberList.append(i)
+        elif len(i) == 1 :
```
fabischw, line 65: Your solution to exercise two adds strings to the charList OR to the numberList. Input "1" is only added to the numberList, not to the string list. The requirements state that the numberList should contain ALL values that can be interpreted as a number and the charList should contain ALL values that are strings of length 1. "1" should therefore end up in both arrays.

fabischw (changes requested):
There are a few small issues as highlighted in my comments.
You may also want to consider to add docstrings to your functions.

fabischw (approved):
requested changes were implemented; looks good. Consider using a different iterator variable next time if it is clear what the objects you are iterating over are - in this example elements or inpt_strings etc. .

PR #43, opened 2023-10-16
Jasmin02

not-too-deranged:


Exercises/JasminSteidl/Übung2.py
```diff
@@ -0,0 +1,89 @@
+# Write a function that takes as input a list of integers and returns a single integer number.
```
paulseidel08, line 1: We were supposed to put the second task in a new branch and not in the same branch as the first.

Exercises/JasminSteidl/Übung2.py
```diff
@@ -0,0 +1,89 @@
...
+#   The first list should contain all arguments which can be interpreted
+#   as a number.
+#   The second list should contain all strings which contain just one character.
+# Think of some good inputs to test this functionality, write down at least three
+# examples and verify that the output for these examples is correct.
+#test1="1","2","3","4"
+#test2="q","q","w","e","r"
+example="1","0","w","2.4","F","G","-6","F","123"
+
+def function(*args):
```
paulseidel08, line 75: better use a function-name that describes what the function does. This is part of comprehensive code.

paulseidel08: Same for the variables

Exercises/JasminSteidl/Übung2.py
```diff
@@ -0,0 +1,89 @@
...
+    randos = []
+    sillies = []
+    
+    for arg in args:
+        if arg.isdigit () or (arg[0]=="-" and arg[1:].isdigit()): #for negative numbers, floats
```
paulseidel08, line 80: depending on the way you interpret the task, that's a very good idea to filter also floats and negative numbers!

paulseidel08: correction: you enable negative numbers, but no floats. Or am I wrong with that? I'm not completely sure.

paulseidel08 (commented):
All in all a very good code. But you shouldn't upload multiple exercises into one file.

Exercises/JasminSteidl/Übung2.py
```diff
@@ -0,0 +1,92 @@
...
+
+def stringFunction(*args):
+    numbers = []
+    singleCharStrings = []
+    
+    for arg in args:
+        if arg.isdigit () or (arg[0]=="-" and arg[1:].isdigit()) \
+            or (arg[0].isdigit() and arg[1]=="." and arg[2].isdigit()) \
+            or (arg[0]=="." and arg[1].isdigit()) \
+            or (arg[0]=="-" and arg[1].isdigit() and arg[2]=="." and arg[3:].isdigit()): #for negative numbers and floats
```
thecodingcube77, line 83: This is a very unreadable if statement. What you are trying to do here is to filter out floats, but you can only filter out numbers with a floating point at the index 1 and no other. So you would not filter out the number `1234.5678`, because it does not fall under any part of your if statement.
I would suggest a cast to float, which is way easier to read and understand. But be aware that a cast to float will throw an error if you give it something that can not e cast to a float.

Exercises/JasminSteidl/Übung2.py
```diff
@@ -0,0 +1,92 @@
...
+# [1, 1, 1, 4, 99, 5, 6, 0, 99] should become [30, 1, 1, 4, 2, 5, 6, 0, 99]
+# Your function should return 30.
+
+#testList = [1,1,1,4,99,5,6,0,99] #for troubleshooting
+
+# print out which value is returned by your function for the following list:
+commands = [1, 12, 2, 3, 1, 1, 2, 3, 1, 3, 4, 3, 1, 5, 0, 3, 2, 1, 9, 19, 1, 5, 19, 23, 1, 6, 23, 27, 1, 27, 10, 31, 1, 31, 5, 35, 2, 10, 35, 39, 1, 9, 39, 43, 1, 43, 5, 47, 1, 47, 6, 51, 2, 51, 6, 55, 1, 13, 55, 59, 2, 6, 59, 63, 1, 63, 5, 67, 2, 10, 67, 71, 1, 9, 71, 75, 1, 75, 13, 79, 1, 10, 79, 83, 2, 83, 13, 87, 1, 87, 6, 91, 1, 5, 91, 95, 2, 95, 9, 99, 1, 5, 99, 103, 1, 103, 6, 107, 2, 107, 13, 111, 1, 111, 10, 115, 2, 10, 115, 119, 1, 9, 119, 123, 1, 123, 9, 127, 1, 13, 127, 131, 2, 10, 131, 135, 1, 135, 5, 139, 1, 2, 139, 143, 1, 143, 5, 0, 99, 2, 0, 14, 0]
+
+def simComputer (input):
+    modInput=input.copy() #input copy not to f with original
```
thecodingcube77, line 34: It is not necessary to copy the array because you don't need it for anything else in this case

Exercises/JasminSteidl/Übung2.py
```diff
@@ -0,0 +1,92 @@
...
+            value2=modInput[modInput[pointer+2]]
+            placeResult=modInput[pointer+3]
+            modInput[placeResult]=value1+value2
+            pointer+=4
+        elif opcode==2:
+            value1=modInput[modInput[pointer+1]]
+            value2=modInput[modInput[pointer+2]]
+            placeResult=modInput[pointer+3]
+            modInput[placeResult]=value1*value2
+            pointer+=4
```
thecodingcube77, line 51: The variables make the code very understandable and easy to follow.

thecodingcube77, line 51: The addition of 4 at the end of both if's  can be done at the end of the for loop so you only have to write this line once. It also helps to prevent bugs in a larger scale. If you want to change the 4 to a 6 for example, you would have to do it twice in your solution, of which you can forget it once. If you only have one place to change the value you cannot have them out of sync.

thecodingcube77 (changes requested):
Your check for numbers should be changed so that it is more consistent

Exercises/JasminSteidl/Übung2.py
```diff
@@ -70,20 +72,21 @@ def simComputer (input):
...
+        if len(arg)==1:
             singleCharStrings.append(arg)
+        try:
+            numbers.append(arbNumbers(arg))
+        except ValueError:
+            pass
+            
+
     return numbers, singleCharStrings
 result=stringFunction(*example)
```
thecodingcube77, line 91: You do not test your function with 3 examples

Exercises/JasminSteidl/Übung2.py
```diff
@@ -30,31 +30,33 @@
 # print out which value is returned by your function for the following list:
 commands = [1, 12, 2, 3, 1, 1, 2, 3, 1, 3, 4, 3, 1, 5, 0, 3, 2, 1, 9, 19, 1, 5, 19, 23, 1, 6, 23, 27, 1, 27, 10, 31, 1, 31, 5, 35, 2, 10, 35, 39, 1, 9, 39, 43, 1, 43, 5, 47, 1, 47, 6, 51, 2, 51, 6, 55, 1, 13, 55, 59, 2, 6, 59, 63, 1, 63, 5, 67, 2, 10, 67, 71, 1, 9, 71, 75, 1, 75, 13, 79, 1, 10, 79, 83, 2, 83, 13, 87, 1, 87, 6, 91, 1, 5, 91, 95, 2, 95, 9, 99, 1, 5, 99, 103, 1, 103, 6, 107, 2, 107, 13, 111, 1, 111, 10, 115, 2, 10, 115, 119, 1, 9, 119, 123, 1, 123, 9, 127, 1, 13, 127, 131, 2, 10, 131, 135, 1, 135, 5, 139, 1, 2, 139, 143, 1, 143, 5, 0, 99, 2, 0, 14, 0]
 
-def simComputer (input):
-    modInput=input.copy() #input copy not to f with original
+def simComputer (Input):
```
thecodingcube77, line 33: Variable names are always lower case

thecodingcube77 (changes requested):
You should take a look at the naming conventions in python.
You also have to test your function in task 2 where you filter the arguments.

thecodingcube77 (approved):
The requested changes got implemented. I don't see any more problems.

TomKnoden:
Everything in your program seems to be working after fixing the issues that were adressed. Well done!

PR #44, opened 2023-10-17
solution for exercise 2

christianrietzsch:


Exercises/ChristianRietzsch/exercises_02.py
```diff
@@ -0,0 +1,76 @@
...
+                reg2 = memory[memory[current+2]]
+                reg3 = memory[current+3]
+                if(element == 1):
+                    memory[reg3] = reg1 + reg2
+                else:
+                    memory[reg3] = reg1 * reg2
+            case 99:
+                return memory[0];
+            case _:
+                print("Error, Illegal Instruction: {memory[current]}")
```
DrKeks1010, line 48: This should be an f string, i suppose

DrKeks1010 (approved):
Overall a great solution.

Be will find out tomorrow, if singel digits like '2' should also be interpreted as strings with just one character or if the program should exit after parsing '2' to a number, so you might need to adjust this.

Exercises/ChristianRietzsch/exercises_02.py
GeFritz: bro lass mal paar reviews übrig

fabischw (changes requested):
Task 2 does not accept the input "1" as both char and number. It only adds the "1" to the number list, not to the char list.

fabischw (approved):
problem has been solved and requested changes implemented. well done.

PR #45, opened 2023-10-17
Jonas Woche02

DrKeks1010:


fabischw (approved):
Overall great solution. Do the \n's and \t's serve any specific purpose in your docstrings ? They seem to be used as regular characters inthe docstrings.

DrKeks1010:
> Overall great solution. Do the \n's and \t's serve any specific purpose in your docstrings ? They seem to be used as regular characters inthe docstrings.

In VSCode they seem to be required to format the docstring in the way I want. But that might only be VSCode intern.

PR #48, opened 2023-10-18
Philipp Kirschke week 2

The-Linux-User:


CorvinAn97:
it was a good idea to consider the hex numbers! :)

Exercises/Philipp Kirschke/exercises_02.py
```diff
@@ -0,0 +1,90 @@
...
+# Your function should return 30.
+
+# print out which value is returned by your function for the following list:
+commands = [1, 12, 2, 3, 1, 1, 2, 3, 1, 3, 4, 3, 1, 5, 0, 3, 2, 1, 9, 19, 1, 5, 19, 23, 1, 6, 23, 27, 1, 27, 10, 31, 1, 31, 5, 35, 2, 10, 35, 39, 1, 9, 39, 43, 1, 43, 5, 47, 1, 47, 6, 51, 2, 51, 6, 55, 1, 13, 55, 59, 2, 6, 59, 63, 1, 63, 5, 67, 2, 10, 67, 71, 1, 9, 71, 75, 1, 75, 13, 79, 1, 10, 79, 83, 2, 83, 13, 87, 1, 87, 6, 91, 1, 5, 91, 95, 2, 95, 9, 99, 1, 5, 99, 103, 1, 103, 6, 107, 2, 107, 13, 111, 1, 111, 10, 115, 2, 10, 115, 119, 1, 9, 119, 123, 1, 123, 9, 127, 1, 13, 127, 131, 2, 10, 131, 135, 1, 135, 5, 139, 1, 2, 139, 143, 1, 143, 5, 0, 99, 2, 0, 14, 0]
+
+
+def intcode(commands):
+    i = 0
+    while i < len(commands):
+        if commands[i] == 1:
```
s231365, line 37: maybe consider using variables, it makes the code easier to read as well as some comments

Exercises/Philipp Kirschke/exercises_02.py
```diff
@@ -0,0 +1,90 @@
...
+# Write a function that takes an arbitrary number of unnamed arguments
+# All inputs will be of type string.
+# the function sould return two lists:
+#   The first list should contain all arguments which can be interpreted
+#   as a number.
+#   The second list should contain all strings which contain just one character.
+# Think of some good inputs to test this functionality, write down at least three
+# examples and verify that the output for these examples is correct.
+
+def checkHex(s):
```
s231365, line 63: please use a different name for the variable. with s it is not clear what you mean

Exercises/Philipp Kirschke/exercises_02.py
```diff
@@ -0,0 +1,90 @@
...
+        if ((char < '0' or char > '9') and (char < 'A' or char > 'F')):
+            return False
+        else: #If char is a valid hex number
+            return True
+
+inputList = ["#", "-", ">", "P", "y", "245", "Hello", "23", "FFFFFF", "Test", "73a", "72", "2457965", "t", "7a", "Python", "py", "o", "-23", "sort", "431", "82ac", "n", "0", "2", "Whitespace", " ", ".", "2.3"]
+number_list = []
+character_list = []
+i = 0
+while i < len(inputList):
```
s231365, line 75: same here, use a different name for the variable. i is not meaningful

s231365 (changes requested):
Overall your solution does what it should, but i don't understand how you check if the inputs are numbers (exercise 2). I only see the check for Hex numbers.

The-Linux-User:
> Overall your solution does what it should, but i don't understand how you check if the inputs are numbers (exercise 2). I only see the check for Hex numbers.

I split all strings in their individual chars. I then check every char if it is a valid hex number (so 1, 2, 3, 4, 5, 6, 7, 8, 9, 0, A, B, C, D, E, F) Non-hex numbers are of course included in the regular numbers from 0-9, so they pass as well. If all chars match the criteria above, the whole number will be added to the number list

Exercises/Philipp Kirschke/exercises_02.py
The-Linux-User: I wrote all of this before he told us on Wednesday not to use i, ... for variables. Will change it later

The-Linux-User: Has been changed to index

Exercises/Philipp Kirschke/exercises_02.py
The-Linux-User: Has been changed to string

Exercises/Philipp Kirschke/exercises_02.py
The-Linux-User: Has been changed to index

s231365 (approved):

PR #50, opened 2023-10-18
exercise_01

CorvinAn97:


The-Linux-User:
Only issue I've noticed is, that the last print only gives out the amount of "a"s in the first list. This works in this specific example but not if one of the lists would change so that they have a different amount of "a"

The-Linux-User (approved):

PR #57, opened 2023-10-19
Lösungen Woche 2

s231365:
bitte reviewen

### DIFF
--- a/Exercises/LauravanUffelt/VL2 Lösungen.py
+++ b/Exercises/LauravanUffelt/VL2 Lösungen.py
@@ -27,27 +27,27 @@
 
 
 # print out which value is returned by your function for the following list:
-commands = [1, 12, 2, 3, 1, 1, 2, 3, 1, 3, 4, 3, 1, 5, 0, 3, 2, 1, 9, 19, 1, 5, 19, 23, 1, 6, 23, 27, 1, 27, 10, 31, 1, 31, 5, 35, 2, 10, 35, 39, 1, 9, 39, 43, 1, 43, 5, 47, 1, 47, 6, 51, 2, 51, 6, 55, 1, 13, 55, 59, 2, 6, 59, 63, 1, 63, 5, 67, 2, 10, 67, 71, 1, 9, 71, 75, 1, 75, 13, 79, 1, 10, 79, 83, 2, 83, 13, 87, 1, 87, 6, 91, 1, 5, 91, 95, 2, 95, 9, 99, 1, 5, 99, 103, 1, 103, 6, 107, 2, 107, 13, 111, 1, 111, 10, 115, 2, 10, 115, 119, 1, 9, 119, 123, 1, 123, 9, 127, 1, 13, 127, 131, 2, 10, 131, 135, 1, 135, 5, 139, 1, 2, 139, 143, 1, 143, 5, 0, 99, 2, 0, 14, 0]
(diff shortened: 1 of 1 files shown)


Exercises/LauravanUffelt/VL2 Lösungen.py
```diff
@@ -0,0 +1,109 @@
...
+
+# Here's another testcase:
+# [1, 1, 1, 4, 99, 5, 6, 0, 99] should become [30, 1, 1, 4, 2, 5, 6, 0, 99]
+# Your function should return 30.
+
+
+# print out which value is returned by your function for the following list:
+commands = [1, 12, 2, 3, 1, 1, 2, 3, 1, 3, 4, 3, 1, 5, 0, 3, 2, 1, 9, 19, 1, 5, 19, 23, 1, 6, 23, 27, 1, 27, 10, 31, 1, 31, 5, 35, 2, 10, 35, 39, 1, 9, 39, 43, 1, 43, 5, 47, 1, 47, 6, 51, 2, 51, 6, 55, 1, 13, 55, 59, 2, 6, 59, 63, 1, 63, 5, 67, 2, 10, 67, 71, 1, 9, 71, 75, 1, 75, 13, 79, 1, 10, 79, 83, 2, 83, 13, 87, 1, 87, 6, 91, 1, 5, 91, 95, 2, 95, 9, 99, 1, 5, 99, 103, 1, 103, 6, 107, 2, 107, 13, 111, 1, 111, 10, 115, 2, 10, 115, 119, 1, 9, 119, 123, 1, 123, 9, 127, 1, 13, 127, 131, 2, 10, 131, 135, 1, 135, 5, 139, 1, 2, 139, 143, 1, 143, 5, 0, 99, 2, 0, 14, 0]
+
+def function(input):
```
LeMixer, line 32: You defined the parameter `input`, but used `commands` in the global scope. You should be using the input parameter instead of global variables inside the function according to the task: "_a function that takes as input a list of integers_"

Exercises/LauravanUffelt/VL2 Lösungen.py
```diff
@@ -0,0 +1,109 @@
...
+# All inputs will be of type string.
+# the function should return two lists:
+#   The first list should contain all arguments which can be interpreted
+#   as a number.
+#   The second list should contain all strings which contain just one character.
+# Think of some good inputs to test this functionality, write down at least three
+# examples and verify that the output for these examples is correct
+
+
+def lists(*arg):
```
LeMixer, line 76: `arg` can be very misleading. Use the plural instead according to the mentioned naming convention.

LeMixer (changes requested):
The code is working alright, but there are some issues. Also consider using docstrings (not mandatory)

Exercises/LauravanUffelt/VL2 Lösungen.py
```diff
@@ -0,0 +1,109 @@
...
+    result1=[]
+    result2=[]
+    for elements in arg:
+        try: result1.append(float(elements))
+        except:
```
thecodingcube77, line 81: You should specify which type of error you want to catch with the `except`. Right now you assume that the solution to any error is to do nothing. In this case I think that you want to catch a `ValueError`.

thecodingcube77 (changes requested):

Exercises/LauravanUffelt/VL2 Lösungen.py
```diff
@@ -73,12 +73,12 @@ def function(input):
...
 
-def lists(*arg):
+def lists(*args):
     result1=[]
     result2=[]
-    for elements in arg:
+    for elements in args:
         try: result1.append(float(elements))
-        except:
+        except (ValueError):
```
thecodingcube77, line 81: The brackets are not needed

LeMixer (approved):

thecodingcube77 (approved):

PR #63, opened 2023-10-20
Week 2 Utku Acikel

UtkuDHBW:
Please review

### DIFF
--- a/Exercises/Utku Acikel/exercises_02.py
+++ b/Exercises/Utku Acikel/exercises_02.py
@@ -1,0 +1,45 @@
+# Write a function that takes as input a list of integers and returns a single integer number.
+# the numbers passed as argument form the working memory of a simulated computer.
+# this computer will start by looking at the first value in the list passed to the function.
+# this value will contain an `opcode`. Valid opcodes are 1, 2 or 99.
+# Encountering any other value when you expect an opcode indicates an error in your coding.
+# Meaning of opcodes:
+#  1 indicates addition. If you encounter the opcode 1 you should read values from two positions 
+#    of your working memory, add them, and store the result in a third position of your working memory.
+#    The three numbers immediately after your opcode indicate the memory locations to read (first two values)
+#    and write (third value) respectively. 
+#    After executing the addition you should move to the next opcode by stepping forward 4 positions.
+#  2 indicates multiplication. Otherwise the same rules apply as for opcode 1.
+# 99 indicates halt. the program should stop after encountering the opcode 99.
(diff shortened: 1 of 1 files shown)


atrejut:
remember to add the label for each week please :)

Exercises/Utku Acikel/exercises_02.py
```diff
@@ -0,0 +1,74 @@
...
+# and ending the program. The final memory looks like [2, 0, 0, 0, 99]. Your function should return 2.
+
+# Here's another testcase:
+# [1, 1, 1, 4, 99, 5, 6, 0, 99] should become [30, 1, 1, 4, 2, 5, 6, 0, 99]
+# Your function should return 30.
+
+
+# print out which value is returned by your function for the following list:
+
+def sim_computer(opcode):
```
LeMixer, line 31: Naming the variable `opcode` can be very misleading since the task describes opcode as the following:

> this (first) value will contain an 'opcode'. Valid opcodes are 1, 2 or 99.

Therefore only every fourth element in the commands list is considered an opcode.
Otherwise the solution looks good.

Exercises/Utku Acikel/exercises_02.py
```diff
@@ -0,0 +1,74 @@
...
+# Think of some good inputs to test this functionality, write down at least three
+# examples and verify that the output for these examples is correct.
+
+def sort_string(*args): # Using *args  for variable number of arguments
+    numbers = []
+    one_char_strings = [] # Initializing two emty lists
+    for arg in args: # Iterating each arguments in args tuple
+        if arg.isnumeric(): # isnumeric : returns True if all characters are numeric in a string
+            numbers.append(arg) # append : Adds Element to the list
+        elif len(arg) == 1:
```
LeMixer, line 63: From my understanding of the task '5' is also a string, which contains just one character, regardless if the the string can be interpreted as a number. In your implementation a string can only either be interpreted as a number *or* be a one char string.

LeMixer (changes requested):

Exercises/Utku Acikel/exercises_02.py
UtkuDHBW: In my understanding, I thought that numbers also with 1 character should be only interpreted as a number. But I think either way is right.

LeMixer (approved):

atrejut:
please only merge after we had a chance to discuss the solutions (in the wednesday class)

UtkuDHBW:
Oh okay sorry, i thougt that we should merge before wednesday.

PR #64, opened 2023-10-21
Gerrit Fritz Woche 3 1306

GeFritz:
1306

### DIFF
--- a/Exercises/GerritFritz/exercises_03.py
+++ b/Exercises/GerritFritz/exercises_03.py
@@ -46,14 +46,12 @@ class SkatDeck(BasicDeck):
 
 fr_deck = FrenchDeck()
 sk_deck = SkatDeck()
-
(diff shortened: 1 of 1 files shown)


Exercises/GerritFritz/exercises_03.py
```diff
@@ -0,0 +1,87 @@
...
+
+def is_valid(digits):
+    has_adjacent = False
+    has_increase = True
+    for i, digit in enumerate(digits):
+        if i > 0 and digit < digits[i-1]:
+            has_increase = False
+        if i < len(digits)-2 and digit == digits[i+1] and digits.count(digit)<=2:
+            has_adjacent = True
+    return has_adjacent and has_increase
```
LeMixer, line 65: This function does not produce the correct output.  It misses a couple hundred numbers. For example `134566` and `135556` are recognized as being not valid, even though they meet the criteria of all digits ascending and at least two same ajacent digits. Line 63 evaluates as false for these numbers.

LeMixer (changes requested):

Exercises/GerritFritz/exercises_03.py
GeFritz: First one is true, but second one is supposed to be wrong. Reread the assignment. Exactly two not more or equal than two.

GeFritz: It works now.

LeMixer: Yeah, you are right about the second one, that was my mistake. However the conversation should no be resolved yet, since your this statement ```assert is_valid([1, 3, 4, 5, 9, 9])``` throws an assertion error when it shouldn't

LeMixer (changes requested):

Exercises/GerritFritz/exercises_03.py
GeFritz: On my machine using that same assert statement there is no such error.

GeFritz: You seem to have pushed another file to my branch.

LeMixer: Sorry, I probably got mixed up with my own code.

LeMixer (approved):
looks good to me now

PR #68, opened 2023-10-23
Jasmin03

not-too-deranged:


CorvinAn97:
you have pushed three documents. But you should push only one document

not-too-deranged:
And if I had even the faintest idea on how to stop that from happening I would absolutely do it. Sorry

atrejut:
> And if I had even the faintest idea on how to stop that from happening I would absolutely do it. Sorry

Not sure how the files ended up in your branch, but I could imagine you're creating new branches from the branch for your last exercise, rather than from the `main` branch? [it matters which branch you're in when you run `git checkout -b` to create a new branch]. in general a good workflow would be:

**before** starting work on a new exercise:
- check out the `main` branch -- `git checkout main`
- make sure you have all remote changes in your local branch -- `git pull`
- create a new branch for the current exercise -- `get checkout -b <branch_name>`
- create a new file for the exercise, write & test the code for your solution
- add the file you created -- `git add <filename>`
- push the changes to the remote -- `git push` (will need some extra options the first time you run it for a new branch, but git will tell you exactly what you need to run)
- create the PR on github.com

hope that helps :)

PR #69, opened 2023-10-23
Jasmin03

not-too-deranged:


Exercises/JasminSteidl/Woche3/Übung3.py
```diff
@@ -0,0 +1,133 @@
...
+        return self.cards[position]
+
+    def __len__(self):
+        return len(self.cards)
+
+    def __iter__(self):
+        return iter(self.cards)
+
+    def __str__(self):
+        return f"{self.rank.name} of {self.suit.symbol}"
```
thecodingcube77, line 31: The class `FrenchCardDeck` does not have the members `rank` and `suit` so would you ever execute this code it would throw an error.

Exercises/JasminSteidl/Woche3/Übung3.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class Rank(FrenchCardDeck):
```
thecodingcube77, line 34: I don't think that you need a class for the ranks, as the rank does not have a specific functionality to it. It just saves the value of the rank. 
It also is not good, that `Rank` is a child class of `FrenchCardDeck`, because a rank is no deck of cards.
The same also applies to the classes `Suit` and `Card`.

Exercises/JasminSteidl/Woche3/Übung3.py
```diff
@@ -0,0 +1,133 @@
...
+
+#import timeit
+
+
+class FrenchCardDeck:
+    def __init__(self):
+        ranks = [Rank(str(number), number) for number in range(2, 11)] + \
+         [Rank('Jack', 11), Rank('Queen', 12), Rank('King', 13), Rank('Ace', 14)]
+        suits = [Suit('Hearts'), Suit('Diamonds'), Suit('Clubs'), Suit('Spades')]
+        self.cards = [Card(rank, suit) for suit in suits for rank in ranks]
```
thecodingcube77, line 19: The initialization of your card deck is quite complicated. I'd suggest you to get away from the classes `Rank`, `Suit` and maybe `Card` to make the whole thing simpler. Your Card class essentially is just a tuple, so a list of tuples should be enough for your cards.

Exercises/JasminSteidl/Woche3/Übung3.py
```diff
@@ -0,0 +1,133 @@
...
+# PART 2:
+# Create a second class that represents a deck of cards usable for Skat -- it should only contain cards from 7 upwards.
+# It should offer all the same functionality of the first class.
+
+class SkatDeck(FrenchCardDeck):
+    def __init__(self):
+        skat_ranks = [Rank(str(value), value) for value in range(7, 11)] + \
+            [Rank('Jack', 11), Rank('Queen', 12), Rank('King', 13), Rank('Ace', 14)]
+        skat_suits = [Suit('Hearts'), Suit('Diamonds'), Suit('Clubs'), Suit('Spades')]
+        self.cards = [Card(rank, suit) for suit in skat_suits for rank in skat_ranks]
```
thecodingcube77, line 71: Here applies the same as for the `FrenchCardDeck` constructor.

Exercises/JasminSteidl/Woche3/Übung3.py
```diff
@@ -0,0 +1,133 @@
...
+# - there is at least one group of exactly two adjacent digits within the number which are the same (like 33 in 123345)
+# - digits only increase going from left to right
+
+def valid_number(digit):
+    str_num = str(digit)
+    has_double = False
+    for index in range(len(str_num) - 1):
+        if str_num[index] == str_num[index + 1] and \
+            (index == 0 or str_num[index] != str_num[index - 1]) and \
+           (index == len(str_num) - 2 or str_num[index] != str_num[index + 2]):
```
thecodingcube77, line 101: What you are trying to do here is to look if there is a pair of two identical digits directly next to each other. This is not really needed in this form. If your digits are in order and a digit is exactly two times in the number, the same digits are automatically next to each other.

thecodingcube77 (changes requested):
The classes in task 1 and 2 are not programmed in a good or correct way. You should think about which classes inherit from which classes. You should only inherit to classes which have an "is an"-relationship to the parent class.
You should overthink your storage solution for the card decks as well.

In task 3 can the code be simplified as stated in the comment, which would make it more understandable. The task 3 is solved well apart from that.

Exercises/JasminSteidl/Woche3/Übung3.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def __iter__(self):
+        return iter(self.cards)
+
+    def __str__(self):
+        return f"{self.rank.name} of {self.suit.symbol}"
+
+
+class Rank(FrenchCardDeck):
+    def __init__(self, name, value):
```
thecodingcube77, line 35: The constructor of a child class should always call the constructor of the parent class with `super().__init__()`. `super()` is a reference to the parent class.

Exercises/JasminSteidl/Woche3/Übung3.py
```diff
@@ -7,83 +7,84 @@
...
-        ranks = [Rank(str(number), number) for number in range(2, 11)] + \
-         [Rank('Jack', 11), Rank('Queen', 12), Rank('King', 13), Rank('Ace', 14)]
-        suits = [Suit('Hearts'), Suit('Diamonds'), Suit('Clubs'), Suit('Spades')]
-        self.cards = [Card(rank, suit) for suit in suits for rank in ranks]
-
-    def __getitem__(self, position):
-        return self.cards[position]
+class CardDeck:
+    def __init__(self, ranks, suits):
+        self.cards = [{'rank': rank, 'suit': suit} for suit in suits for rank in ranks]
```
thecodingcube77, line 20: This is a very readable and simple way to initialize your card deck.

Exercises/JasminSteidl/Woche3/Übung3.py
```diff
@@ -7,83 +7,84 @@
...
+            return self.cards[position]
+        else:
+            raise IndexError("Index out of range")
+
     def __iter__(self):
         return iter(self.cards)
 
     def __str__(self):
-        return f"{self.rank.name} of {self.suit.symbol}"
+        return f"\nYou have a {self.__class__.__name__} with {len(self)} cards"
```
thecodingcube77, line 35: Nice Way to represent your card deck. I didn't even think about using `self.__class__.__name__`.

Exercises/JasminSteidl/Woche3/Übung3.py
```diff
@@ -7,83 +7,84 @@
...
+# Accessing cards by index
+print("This is your first card:")
+card_index = 0
+specific_card = french_deck[card_index]
+print(french_deck.card_description(specific_card))
 
+# Showing all cards
+print("\nThis is the deck you have:")
+for card in french_deck:
+    print(french_deck.card_description(card))
```
thecodingcube77, line 68: This is a way to test the functionality of your deck. It will require you to read the output to verify that it is correct. You could try using `assert` suggested in the task. This is not needed as the task does not require you to use `assert`.

thecodingcube77 (approved):
You fixed the named issues and made the code very readable and easy to understand. Well done

PR #70, opened 2023-10-23
Lösungen Woche 3

s231365:
please review
Task 3: 1306

### DIFF
--- a/Exercises/LauravanUffelt/VL3 Lösungen.py
+++ b/Exercises/LauravanUffelt/VL3 Lösungen.py
@@ -1,0 +1,140 @@
+# PART 1:
+# Write a class for a French deck of cards (2-Ace of diamonds, hearts, spades, clubs).
+# The deck of cards should behave like a sequence.
+# When initialized the cards should all be in a well-defined order (2-Ace of each suite, suites in the order above
+# I should be able to index into the deck to retrieve one card.
+# I should be able to iterate over all cards in the deck.
+# Printing a cards string representation should give me a nice,
+# readable description of that card.
+
+class Cards():
+    def __init__(self, number, symbol):
+        self.symbol = symbol
+
+        if number >= 2 and number <= 10:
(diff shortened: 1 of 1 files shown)


Exercises/LauravanUffelt/VL3 Lösungen.py
```diff
@@ -88,53 +88,21 @@ def count(a, b):
...
+            f = digits.count(6)
+            g = digits.count(7)
+            h = digits.count(8)
+            i = digits.count(9)
+            if a == 2 or b == 2 or c == 2 or d == 2 or e == 2 or f == 2 or g == 2 or h == 2 or i == 2:
+                print(number)
+                counter += 1
         else:
             continue
+    print(counter)
```
DrKeks1010, line 105: Your function should return the count, not just print it.
This is the only mistake, that should be changed

Exercises/LauravanUffelt/VL3 Lösungen.py
```diff
@@ -88,53 +88,21 @@ def count(a, b):
...
+            b = digits.count(2)
+            c = digits.count(3)
+            d = digits.count(4)
+            e = digits.count(5)
+            f = digits.count(6)
+            g = digits.count(7)
+            h = digits.count(8)
+            i = digits.count(9)
+            if a == 2 or b == 2 or c == 2 or d == 2 or e == 2 or f == 2 or g == 2 or h == 2 or i == 2:
+                print(number)
```
DrKeks1010, line 101: you may change it, that the valid number is added to an array, which gets printed in the end. That way you can still see the feedback of the first two excercises after this function executed. My command line didnt't cache more than 1306 lines.

DrKeks1010 (changes requested):
Great solution overall.

Please change the last function to also return the number of elements found.

DrKeks1010 (approved):
Looks Great.
All Issues fixed

PR #76, opened 2023-10-25
Philipp Kirschke exercise03.py completed

The-Linux-User:
Count of numbers matching criteria: 1306


Exercises/Philipp Kirschke/exercises_03.py
```diff
@@ -0,0 +1,112 @@
...
+    number = str(number)
+    for i in range(len(number)-1):
+        if (number[i] > number[i+1]):
+            return False
+    return True
+
+def checkDouble(number):
+    number = str(number)
+    for index in range(len(number)-1):
+        if (number[index] == number[index+1] and number.count(number[index]) == 2):
```
TrueEsel, line 105: since the numbers checked here are already sorted number[index] == number[index+1] is unnecessary. The right part of the and statement for itself yields the same result.

Exercises/Philipp Kirschke/exercises_03.py
```diff
@@ -0,0 +1,112 @@
...
+class Card:
+    def __init__(self, value, suit):
+        self.value = value
+        self.suit = suit
+        self.name = str(value) + " of " + suit
+
+    def __str__(self):
+        return self.name
+
+class Deck:
```
TrueEsel, line 20: In my opinion Deck is not a good name for this class since the skat class represents a card deck too

Exercises/Philipp Kirschke/exercises_03.py
```diff
@@ -0,0 +1,112 @@
...
+elif (int(cardIndex) > 51):
+    print("There are only 52 cards in a deck.")
+else:
+    print("The card at index " + cardIndex + " is a " + str(deck.getcard(int(cardIndex))))
+
+
+# PART 2:
+# Create a second class that represents a deck of cards usable for Skat -- it should only contain cards from 7 upwards.
+# It should offer all the same functionality of the first class.
+class skat:
```
TrueEsel, line 53: classes should start with an uppercase letter

Exercises/Philipp Kirschke/exercises_03.py
```diff
@@ -0,0 +1,112 @@
...
+                self.cards.append(Card(value, suit))
+
+    def getcard(self, index):
+        return self.cards[index]
+
+    def iterate(self):
+        return iter(self.cards)
+
+
+# Write some code to test the functionality of both kinds of decks. (You can use `assert` to make sure your classes behave the way you expect them to.)
```
TrueEsel, line 70: the test cases for the Skat class are missing

Exercises/Philipp Kirschke/exercises_03.py
```diff
@@ -0,0 +1,112 @@
...
+    def getcard(self, index):
+        return self.cards[index]
+
+    def iterate(self):
+        return iter(self.cards)
+
+
+
+deck = Deck()
+cardIndex = input("Enter the index of the card you want to see or press enter to see all cards in the deck: ")
```
TrueEsel, line 39: nice test case, that lets the user pick the card

Exercises/Philipp Kirschke/exercises_03.py
```diff
@@ -0,0 +1,112 @@
...
+# - 111334 is a valid number. while there are three 1s, there is also a group of exactly two 3s.
+# - 112233 is a valid number. At least one group of two is fulfilled, there is no maximum to the number of such groups.
+
+def checkNumber(lowerBound, upperBound):
+    numberCount = 0
+    for number in range(lowerBound, upperBound):
+        if(checkIncrease(number)):
+            if(checkDouble(number)):
+                numberCount += 1
+    print(numberCount)
```
TrueEsel, line 93: checkNumber should return the number, not just print it

Exercises/Philipp Kirschke/exercises_03.py
```diff
@@ -0,0 +1,112 @@
...
+
+def checkIncrease(number):
```
TrueEsel, line 95: interesting way to check if the number is increasing from left to right

TrueEsel (changes requested):
Please add the test cases for the Skat deck and the return value for the third task. Its nothing major but still part of the tasks, should be an easy fix :)

Exercises/Philipp Kirschke/exercises_03.py
TrueEsel: you could also let skat inherit from deck since its a special case of the french deck but your code works too

Exercises/Philipp Kirschke/exercises_03.py
The-Linux-User: Removed the redundant code

Exercises/Philipp Kirschke/exercises_03.py
The-Linux-User: Changed the name to FrenchDeck

Exercises/Philipp Kirschke/exercises_03.py
The-Linux-User: Changed the class name to uppercase, this was a typo

Exercises/Philipp Kirschke/exercises_03.py
The-Linux-User: Added the same user defined test cases that the FrenchDeck has

Exercises/Philipp Kirschke/exercises_03.py
The-Linux-User: Thanks!

Exercises/Philipp Kirschke/exercises_03.py
The-Linux-User: Added return statement

Exercises/Philipp Kirschke/exercises_03.py
The-Linux-User: Could've also done it the other way around, but while having the flu and fever this seemed like the better method, I don't know why XD

Exercises/Philipp Kirschke/exercises_03.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def getcard(self, index):
+        return self.cards[index]
+
+    def iterate(self):
+        return iter(self.cards)
+
+
+# Write some code to test the functionality of both kinds of decks. (You can use `assert` to make sure your classes behave the way you expect them to.)
+scatdeck = FrenchDeck()
```
TrueEsel, line 71: your variable scatdeck is an object of FrenchDeck, not of Skat.

Exercises/Philipp Kirschke/exercises_03.py
```diff
@@ -0,0 +1,122 @@
...
+cardIndex = input("Enter the index of the Skat card you want to see or press enter to see all cards in the deck: ")
+if (cardIndex == ""):
+    scatdeck.iterate()
+    for card in scatdeck.iterate():
+        print(card)
+elif (int(cardIndex) > 51):
```
TrueEsel, line 77: since a skat deck has less cards than a french deck this condition is no longer working how its supposed to. If I search for the card at index 51 this will give me an IndexError

TrueEsel (changes requested):
Code is looking good, aside from the mistakes in the Skat Test. Fix that and your finished :)

Exercises/Philipp Kirschke/exercises_03.py
The-Linux-User: Both of the issues should be fixed now

TrueEsel (approved):
looks good now :)

PR #77, opened 2023-10-25
Paul seidel exercise03

paulseidel08:
Task 03 is ready to be reviewed. I hope, theres nothin doubled now.
Task 3.3 returns 331261


joh-lin (commented):
All in all I really like how you implemented the card deck. The sequence-like behaviour for your was correctly implemented and commented. 
A possible improvement for your code would be to make both french_card_deck and skat_deck inherit from a base class, so you could reuse code. The functionality of your sequence behaviour could also have been tested more thoroughly.
 For future projects please remember that class names should be written in pascal case like "Card" or "FrenchCardDeck".

Exercises/PaulSeidel/PSeidel_exercises_03.py
```diff
@@ -0,0 +1,124 @@
...
+        num = []
+        for j in range(10): # creating a list that saves every number's count
+            j_as_str = str(j)
+            num.append(temp.count(j_as_str))
+        if num.count(2) != 0:
+            counter += 1
+    return counter
+
+
+print(str(num_counter(min = 134564, max = 585159))) # finds 331261 numbers
```
joh-lin, line 124: Your solution for part 3 doesn't do exactly what the task requires. It also returns an incorrect solution value.

joh-lin (changes requested):
I liked part 1 and 2. Part 3 needs some fixing.

Exercises/PaulSeidel/PSeidel_exercises_03.py
paulseidel08: To-Dos for myself:
- It's not checking, if the numbers are next to each other (there should be at least two next to each other)
- It's not checking, if the numbers are sorted

joh-lin: Try adding the check if they are sorted first. It might already solve the problem.

paulseidel08: 👍

joh-lin (approved):

PR #78, opened 2023-10-25
Jan Merz Exercise 03

TrueEsel:
Return value of task 3: 1306

### DIFF
--- a/Exercises/JanMerz/exercises_03.py
+++ b/Exercises/JanMerz/exercises_03.py
@@ -17,7 +17,7 @@ class Card:
     def __str__(self):
         return f"{self.rank} of {self.suit}"
     
-class French_Deck:
+class FrenchDeck:
     suits = ["Diamonds", "Hearts", "Spadess", "Clubs"]
     ranks = ["2", "3", "4", "5", "6", "7", "8", "9", "10", "Jack", "Queen", "King", "Ace"]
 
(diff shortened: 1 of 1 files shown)


Exercises/JanMerz/exercises_03.py
```diff
@@ -0,0 +1,107 @@
...
+            for rank in self.ranks[fromRank:]:
+                self.cards.append(Card(rank, suit))
+
+    def __str__(self):
+        return f", ".join([str(card) for card in self.cards])
+    
+    def __iter__(self):
+        return (card for card in (self.cards))
+
+    def __get__(self, index):
```
LeMixer, line 36: This does not implement indexing. However the task states you should be able to index the card deck to receive a card in the following manner: `card_at_index = deck[index]`
Try using __getitem__ instead.

Exercises/JanMerz/exercises_03.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __init__(self, rank, suit):
+        self.rank = rank
+        self.suit = suit
+    
+    def __str__(self):
+        return f"{self.rank} of {self.suit}"
+    
+class FrenchDeck:
+    suits = ["Diamonds", "Hearts", "Spadess", "Clubs"]
```
LeMixer, line 21: Spades

Exercises/JanMerz/exercises_03.py
```diff
@@ -0,0 +1,107 @@
...
+
+class SkatDeck(FrenchDeck):
+    def __init__(self):
+        super().__init__(5)
+
+
+# Write some code to test the functionality of both kinds of decks. (You can use `assert` to make sure your classes behave the way you expect them to.)
+french_deck = FrenchDeck()
+
+print(french_deck)
```
LeMixer, line 53: Not sure why you are printing the deck twice. The second print statement shows that the deck is in fact iterable, but the first print just clutters the console.

LeMixer (changes requested):
Looks good in general. Please revisit the lines I've commented:

Exercises/JanMerz/exercises_03.py
TrueEsel: Fixed the Typo

Exercises/JanMerz/exercises_03.py
TrueEsel: implemented indexing. Thanks for the tip :)

LeMixer (approved):
Looks Good

PR #82, opened 2023-10-26
Final Exercise_03

CorvinAn97:
Solution of Task 3.3 is 1306

### DIFF
--- a/Exercises/Corvin Ansorge/exercises_03.py
+++ b/Exercises/Corvin Ansorge/exercises_03.py
@@ -1,0 +1,139 @@
+# PART 1:
+# Write a class for a French deck of cards (2-Ace of diamonds, hearts, spades, clubs).
+# The deck of cards should behave like a sequence.
+# When initialized the cards should all be in a well-defined order (2-Ace of each suite, suites in the order above
+# I should be able to index into the deck to retrieve one card.
+# I should be able to iterate over all cards in the deck.
+# Printing a cards string representation should give me a nice, 
+# readable description of that card.
+import select
+
+#print("\nTask 1\n")
+
+class Card:
+    def __init__(self, value, suit):
+        self.value = value
+        self.suit = suit
+
+    def __repr__(self):
+        return f"{self.value} of {self.suit}"
+
+
+class FrenchDeck:
+    def __init__(self):
+        self.cards = []
+        self.suits = ["diamonds", "hearts", "spades", "clubs"]
+        self.values = ["2", "3", "4", "5", "6", "7", "8", "9", "10",
+                       "Jack", "Queen", "King", "Ace"]
+        for suits in self.suits:
+            for values in self.values:
+                self.cards.append(Card(values, suits))
+
+    def __getitem__(self, position):
+        return self.cards[position]
+
+
+
+class ScatDeck(FrenchDeck):
+    def __init__(self):
+        super().__init__()
+        self.cards = [card for card in self.cards if card.value in [
+            "7", "8", "9", "10", "Jack", "Queen", "King", "Ace"]]
+
+
+frenchdeck = FrenchDeck()
+scatdeck = ScatDeck()
+
+
+# Write some code to test the functionality of both kinds of decks. (You can use `assert` to make sure your classes behave the way you expect them to.)
+
+def test_frenchdeck():
+    assert len(frenchdeck.cards) == 52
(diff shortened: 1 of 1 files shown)


The-Linux-User (changes requested):
The count of numbers from task 3 should be mentioned in your pull request

CorvinAn97:
> The count of numbers from task 3 should be mentioned in your pull request

fixed that. :)

The-Linux-User (approved):
Requested changes implemented

Exercises/Corvin Ansorge/exercises_03.py
```diff
@@ -0,0 +1,139 @@
+# PART 1:
+# Write a class for a French deck of cards (2-Ace of diamonds, hearts, spades, clubs).
+# The deck of cards should behave like a sequence.
+# When initialized the cards should all be in a well-defined order (2-Ace of each suite, suites in the order above
+# I should be able to index into the deck to retrieve one card.
+# I should be able to iterate over all cards in the deck.
+# Printing a cards string representation should give me a nice, 
+# readable description of that card.
+import select
```
thecodingcube77, line 9: This import is never used and should be removed.

Exercises/Corvin Ansorge/exercises_03.py
```diff
@@ -0,0 +1,139 @@
...
+
+def test_frenchdeck():
+    assert len(frenchdeck.cards) == 52
+    assert frenchdeck.cards[0] == Card("2", "diamonds")
+    assert frenchdeck.cards[51] == Card("Ace", "clubs")
+    assert frenchdeck.cards[12] == Card("Ace", "diamonds")
+    assert frenchdeck.cards[13] == Card("2", "hearts")
+    assert frenchdeck.cards[25] == Card("Jack", "spades")
+    assert frenchdeck.cards[38] == Card("Queen", "clubs")
+    assert frenchdeck.cards[51] == Card("Ace", "clubs")
```
thecodingcube77, line 58: I think the task requires us to implement sequence behavior to the CardDeck classes so that you can index the CardDeck-object directly like this: `frenchdeck[0]`. This should result in `Card("2", "diamonds")` in your case.

Exercises/Corvin Ansorge/exercises_03.py
```diff
@@ -0,0 +1,139 @@
...
+    for num in range(lower_bound, upper_bound):
+        num_str = str(num)
+        has_adjacent = False
+        increasing = True
+
+        for i in range(len(num_str) - 1):
+            if num_str[i] > num_str[i + 1]:
+                increasing = False
+                break
+            if num_str[i] == num_str[i + 1] and (i == 0 or num_str[i - 1] != num_str[i]) and (i + 2 == len(num_str) or num_str[i + 2] != num_str[i]):
```
thecodingcube77, line 130: This if is logically completely fine as it does exactly what you need. But it is quite an overcomplication in my opinion. When you think about it, you don't actually need to verify that the group of 2 digits has no third digit in front or behind it. This is because all the same digits are next to each other when the digits are in order, so you only have to test if there is one digit that is there exactly 2 times in your number.

thecodingcube77 (commented):
Solid solution. Very much blank lines for some reason.

Exercises/Corvin Ansorge/exercises_03.py
CorvinAn97: Thanks for the reminder!

Exercises/Corvin Ansorge/exercises_03.py
CorvinAn97: Changed that! :)

Exercises/Corvin Ansorge/exercises_03.py
CorvinAn97: Good point. I also changed there a bit

Exercises/Corvin Ansorge/exercises_03.py
```diff
@@ -0,0 +1,106 @@
...
+
+#print(frenchdeck.cards)
+#print(scatdeck.cards)
+
+def check_number(number):
+    number = str(number)
+    if number != "".join(sorted(number)):
+        return False
+    for i in range(10):
+        if str(i)*2 in number and str(i)*3 not in number:
```
thecodingcube77, line 95: That is a very interesting way to test for the pair of digits

PR #84, opened 2023-10-26
Add files via upload

HenrietteNeumann:
Aufgaben zu Woche 3

### DIFF
--- a/exercise._03.py
+++ b/exercise._03.py
@@ -1,0 +1,143 @@
+# PART 1:
+# Write a class for a French deck of cards (2-Ace of diamonds, hearts, spades, clubs).
+# The deck of cards should behave like a sequence.
+# When initialized the cards should all be in a well-defined order (2-Ace of each suite, suites in the order above
+# I should be able to index into the deck to retrieve one card.
+# I should be able to iterate over all cards in the deck.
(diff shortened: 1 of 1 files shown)


Reisii:
Nice solution overall. No quite sure if the solution for the last task will be working though

atrejut:
@HenrietteNeumann please try to get this old PR merged...

exercise._03.py
```diff
@@ -0,0 +1,145 @@
...
+    count = 0
+
+    for number in range(lower_bound, upper_bound):
+        num_str = str(number)
+        has_adjacent = False
+        is_increasing = True
+
+        for i in range(len(num_str) - 1):
+            if num_str[i] == num_str[i + 1]:
+                has_adjacent = True
```
fabischw, line 128: the problem with this code snippet is that it only checks if there is the same digit next to the current one. If there is 3 adjacent digits (ex: 113335) this will still be set to True although the task specifies that only EXACTLY 2 adjacent digits should be counted.

fabischw (changes requested):
Result for week 2 is off as pointed out in the comment. Please also move this file as it is currently in the wrong directory.

PR #87, opened 2023-10-26
Sorry for late submission. Had some issues with Internet :(

Reisii:


Exercises/TuNguyenXuan/Tufgaben3.py
```diff
@@ -0,0 +1,110 @@
+# PART 1:
+# Write a class for a French deck of cards (2-Ace of diamonds, hearts, spades, clubs).
+# The deck of cards should behave like a sequence.
+# When initialized the cards should all be in a well-defined order (2-Ace of each suite, suites in the order above
+# I should be able to index into the deck to retrieve one card.
+# I should be able to iterate over all cards in the deck.
+# Printing a cards string representation should give me a nice,
+# readable description of that card.
+
+class Card:
```
paulseidel08, line 10: Remember to use Docstrings (These Comments with """...""" after the initialization of functions, classes and methods, to make it easier to use those later on.

Exercises/TuNguyenXuan/Tufgaben3.py
```diff
@@ -0,0 +1,110 @@
...
+            for val in self.val[value:]:
+                self.cards.append(Card(val, suit))
+
+    def __str__(self):
+        return f", ".join([str(card) for card in self.cards])
+
+    def __iter__(self):
+        return(card for card in (self.cards))
+
+    def __get__(self, i):
```
paulseidel08, line 37: __get__ is no real dunder method. Use __getitem__ instead.

Exercises/TuNguyenXuan/Tufgaben3.py
```diff
@@ -0,0 +1,110 @@
...
+        return self.cards[i]
+
+
+# PART 2:
+# Create a second class that represents a deck of cards usable for Skat -- it should only contain cards from 7 upwards.
+# It should offer all the same functionality of the first class.
+class SkatCardDeck(FrenchCardDeck):
```
paulseidel08, line 44: It makes sense to shorten that in here, cause the classes do the same thing. The logical problem is, that the SkatCardDeck now is dependent on the FrenchCardDeck, which it shouldn't be. In later codes, better implement a higher class (in this context it that would be "Deck") and then inherit the subclasses (here SkatCardDeck and FrenchCardDeck) to avoid writing the same code twice.

Exercises/TuNguyenXuan/Tufgaben3.py
```diff
@@ -0,0 +1,110 @@
...
+#
+# Examples:
+# - 123345 is a valid number
+# - 123341 is not a valid number, as the digits do not increase from left to right
+# - 123334 is not a valid number as there is no group of exactly two repeated digits
+# - 111334 is a valid number. while there are three 1s, there is also a group of exactly two 3s.
+# - 112233 is a valid number. At least one group of two is fulfilled, there is no maximum to the number of such groups.
+#
+# run your function with the lower bound `134564` and the upper bound `585159`. Note the resulting count
+# in your pull request, please.
```
paulseidel08, line 81: You didn't note your resulting count in you pull request.

Exercises/TuNguyenXuan/Tufgaben3.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def number_filter(lowerbound, upperbound):
+    count = 0
+    numbers = [*range(lowerbound, upperbound)]
+    arr = [str(x) for x in numbers]
```
paulseidel08, line 87: U don't need a list for the numbers between lowerbound and upperbound. To make your code easier and more efficient, you could just use "arr = [str(x) for x in range(lowerbound, upperbound]". That would automatically include lowerbound and exclude upperbound and would make the list "numbers" obsolete.

Exercises/TuNguyenXuan/Tufgaben3.py
```diff
@@ -0,0 +1,110 @@
...
+            else:
+                if arr[i][j] < arr[i][j+1]:
+                    valid_number = True
+                    if arr[i][j] == arr[i][j+1]:
+                        if arr[i][j] != arr[i][j+2]:
+                            valid_number = True
+                else:
+                    valid_number = False
+
+        if valid_number:
```
paulseidel08, line 102: There's a statement missing. Like this, it prints the number everytime the for-loop in line 88 executes.

Exercises/TuNguyenXuan/Tufgaben3.py
```diff
@@ -0,0 +1,110 @@
...
+    count = 0
+    numbers = [*range(lowerbound, upperbound)]
+    arr = [str(x) for x in numbers]
+    for i in range(len(arr)):
+        valid_number = None
+        for j in range(len(arr[i])):
+            if j == len(arr[i])-1:
+                break
+            else:
+                if arr[i][j] < arr[i][j+1]:
```
paulseidel08, line 94: Why is the list "arr" 2-dimensional now? It only needs one parameter.

paulseidel08 (changes requested):
Your solutions for the first two tasks are working good. You could make some improvements, but that's not necessary.
You should use docstrings and comments to make your code more understandable. Please remember that for the next exercises.
Task 3 doesn't work how it should. It prints the numbers over 300000 times and prints the wrong amount of filtered numbers between the bounds.

atrejut:
@Reisii please try to get this old PR merged...

PR #88, opened 2023-10-27
Woche03 AUfgaben

TomKnoden:
There are changes to be made to the file

### DIFF
--- a/Exercises/Tom_Knoden/Woche03/Woche_03_Aufgaben.py
+++ b/Exercises/Tom_Knoden/Woche03/Woche_03_Aufgaben.py
@@ -1,0 +1,84 @@
+#Part 1
+suits = ["diamonds", "hearts", "spades", "clubs"]
+
+class card:
+    def __init__(self, value, suit):
+        self.value = value
+        self.suit = suit
+
+    def __str__(self):
+        return (f"{self.value} of {self.suit}, ")
+
+class french_card_deck:
+    def __init__(self):
+        values = ["2", "3", "4", "5", "6", "7", "8", "9", "10", "jack", "queen", "king", "ace"]
+        self.deck = []
+        for suit in suits:
+            for value in values:
+                self.deck.append(card(value, suit))
+
+    def __getitem__(self, index):
+        if index >= 0 and index < len(self.deck):
+            return self.deck[index]
+        else:
+            raise IndexError("Index out of bound")
+
+    def __iter__(self):
+        return (element for element in self.deck)
+
+    def __str__(self):
+        storage = ""
+        for counter in range(len(self.deck)):
+            storage += str(self.deck[counter])
+        return storage
+
+
+#Part 2
+class skat_deck:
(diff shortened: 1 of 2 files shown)


Exercises/Tom_Knoden/Woche03/Woche_03_Aufgaben.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __iter__(self):
+        return (element for element in self.deck)
+
+    def __str__(self):
+        storage = ""
+        for counter in range(len(self.deck)):
+            storage += str(self.deck[counter])
+        return storage
+
```
s231365, line 34: maybe you could do some testcases to test the functionality of your code

Exercises/Tom_Knoden/Woche03/Woche_03_Aufgaben.py
```diff
@@ -0,0 +1,84 @@
...
+def count_valid_numbers(lower_bound, upper_bound):
+    count = 0
+    for number in range(lower_bound, upper_bound):
+        num_str = str(number)
+        has_adjacent_duplicates = False
+        is_increasing = True
+
+        for i in range(len(num_str) - 1):
+            if num_str[i] == num_str[i + 1]:
```
s231365, line 70: you also include numbers like 579999, which is incorrect. there should only be two 9's like 567899.  You need to check the third digits also.

Exercises/Tom_Knoden/Woche03/Woche_03_Aufgaben.py
```diff
@@ -0,0 +1,84 @@
...
+    def __getitem__(self, index):
+        if index >= 0 and index < len(self.deck):
+            return self.deck[index]
+        else:
+            raise IndexError("Index out of bound")
+
+    def __iter__(self):
+        return (element for element in self.deck)
+
+    def __str__(self):
```
s231365, line 29: this not a string representation. it gives the storage position. since this is a dunder method you should use __str__ for string representation.

Exercises/Tom_Knoden/Woche03/Woche_03_Aufgaben.py
```diff
@@ -0,0 +1,84 @@
...
+        storage = ""
+        for counter in range(len(self.deck)):
+            storage += str(self.deck[counter])
+        return storage
+
+
+#Part 2
+class skat_deck:
```
s231365, line 37: Here you could use inheritance. this would shorten your code, because you sont need to write the funtions twice

s231365 (changes requested):
Good ideas, but you need to make some changes, because in number 3 your result is wrong.

atrejut:
@TomKnoden please try to get this old PR merged...

fabischw (changes requested):
Please remove the tasks for week 4 from this PR and address changes requested by @s231365 .

PR #90, opened 2023-10-27
solution for exercise 4 (16694270)

christianrietzsch:


Exercises/ChristianRietzsch/exercises_04.py
```diff
@@ -0,0 +1,159 @@
...
+            case "06" :
+                reg1 = memory[counter+1] if (len(mode) >= 1 and mode[-1] == "1") else memory[memory[counter+1]]
+                reg2 = memory[counter+2] if (len(mode) >= 2 and mode[-2] == "1") else memory[memory[counter+2]]
+                if(reg1 == 0):
+                    counter  = reg2
+                else:
+                    counter += 3
+            case "07" :
+                reg1 = memory[counter+1] if (len(mode) >= 1 and mode[-1] == "1") else memory[memory[counter+1]]
+                reg2 = memory[counter+2] if (len(mode) >= 2 and mode[-2] == "1") else memory[memory[counter+2]]
```
GeFritz, line 126: Good use of `match case`. The part were `reg1` and `reg2` are assigned could be inserted before the match case. This would make the code less repetitive.

Exercises/ChristianRietzsch/exercises_04.py
```diff
@@ -0,0 +1,159 @@
...
+
+def execute(memory: [int], debug=0, debug_input=0):
```
GeFritz, line 75: Skipping the input is a nice idea to test the code.

GeFritz (approved):
In general good and functional code. Slightly unreadable due to everything being in one large function. Some parts could be excluded into helper functions to help readability. The match case could also be exchanged for a opcode map with functions, this is not necessary but helps readability.

PR #92, opened 2023-10-27
Bo Wichtendahl Week 04 (result: 16694270)

thecodingcube77:
Solved exercises_04.py, resulting number: 16694270


LeMixer (approved):
This is a really good solution in my opinion, well done!

PR #95, opened 2023-10-28
Gerrit Fritz Aufgabe 4 16694270

GeFritz:


Exercises/GerritFritz/exercises_04.py
```diff
@@ -0,0 +1,79 @@
...
+def less(i, pos, commands):
+    if commands[pos[0]] < commands[pos[1]]: commands[pos[2]] = 1
+    else: commands[pos[2]] = 0
+    return i + 4 
+    
+def equals(i, pos, commands):
+    if commands[pos[0]] == commands[pos[1]]: commands[pos[2]] = 1
+    else: commands[pos[2]] = 0
+    return i + 4
+
```
FritzRa, line 38: Good choice to use subfunctions (for every operation) which return the increase of index. This greatly increases the readability. Variable names and function names are also simple and well chosen

Exercises/GerritFritz/exercises_04.py
```diff
@@ -0,0 +1,79 @@
...
+            positions.append(commands[index+i+1])
+        elif mode[i] == "1" and len(commands)>index+i+1:
+            positions.append(index+i+1)
+        else: raise KeyError("Wrong mode")
+    return positions
+
+
+def get_commands(number):
+    command_tuple = divmod(number,100)
+    return command_tuple[1], f"{command_tuple[0]:03d}"[::-1]
```
FritzRa, line 65: interesting way to convert the given command into its opcode and modes as an integer. This maybe could have been done simpler, for instance by converting to a string and splicing. However it is still a completely valid approach.

FritzRa (approved):
Overall great solution to the task. It is readable and functional using some unique approaches to the task.

PR #96, opened 2023-10-29
Raphael Fritz Woche 4 Aufgaben 16694270

FritzRa:


The-Linux-User (changes requested):
Variables should be easy to read, please change your variables (e.g. i) to be easier to read.
Rest looks good to me, interesting solution to use individually defined functions for the opcodes, makes the code nice to read!

FritzRa:
> Variables should be easy to read, please change your variables (e.g. i) to be easier to read. Rest looks good to me, interesting solution to use individually defined functions for the opcodes, makes the code nice to read!

I would be reluctant to change specifically "i" to something else as it is a name convention in python for index. What other variables should i make easier to read?

The-Linux-User:
> > Variables should be easy to read, please change your variables (e.g. i) to be easier to read. Rest looks good to me, interesting solution to use individually defined functions for the opcodes, makes the code nice to read!
> 
> I would be reluctant to change specifically "i" to something else as it is a name convention in python for index. What other variables should i make easier to read?

The others are fine, I like using i as well but atrejut meant two weeks ago that we should avoid variable names like i and use more descriptive names like index. Since it is his Repo and he gives the grades I would change them even though I will continue using shorter variable names in my own projects

FritzRa:
changed i to index

The-Linux-User (approved):

PR #97, opened 2023-10-29
Jan Merz Exercise 04 (16694270)

TrueEsel:


not-too-deranged (approved):
Works with all test cases, docstrings explain what is going on even if code is rather long. The only thing I have noticed is that there is no special error message if there is a wrong opcode, it just raises a keyerror (don't think that's important tho)

paulseidel08 (approved):
Very good and easy readable code. Additionally you could add docstrings for even the smaller functions, explaining which arguments they need to work in which way. But that's not too important.

PR #101, opened 2023-11-01
PSeidel Exercise04 unfinished

paulseidel08:
Had some problems with troubleshooting. My code does not work like it should. A review would really help me.

### DIFF
--- a/Exercises/PaulSeidel/PSeidel_exercises_04.py
+++ b/Exercises/PaulSeidel/PSeidel_exercises_04.py
@@ -1,0 +1,115 @@
+# Extend the simulated computer from the second week:
+# You will need to support a number of additional opcodes:
+# - 3: read a single integer as input and save it to the position given
+#      by its only parameter. the command 3,19 would read an input
+#      and store the result at address 19
+# - 4: output the value of the single parameter for this opcode.
+#      for example 4,19 would output the value stored at address 19
+# - 5: jump-if-true: if the first parameter is non-zero, it sets the instruction pointer to the value from the second parameter. Otherwise, it does nothing.
+# - 6: jump-if-false: if the first parameter is zero, it sets the instruction pointer to the value from the second parameter. Otherwise, it does nothing.
+# - 7: less than: if the first parameter is less than the second parameter, it stores 1 in the position given by the third parameter. Otherwise, it stores 0.
+# - 8: equals: if the first parameter is equal to the second parameter, it stores 1 in the position given by the third parameter. Otherwise, it stores 0.
+#
+# caution: since these opcodes expect a variable number of parameters the instruction pointer after executing an instruction should no longer always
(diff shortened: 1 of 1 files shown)


Exercises/PaulSeidel/PSeidel_exercises_04.py
```diff
@@ -0,0 +1,115 @@
...
+# Immediate mode: In immediate mode a parameter is directly interpreted as a value. If the first argument to the 'sum' opcode is 8
+#   then the first summand in your calculation is 8.
+#
+# Parameter modes are specified per-parameter as part of the opcode by extending the opcodes.
+# When reading a number that specifies an opcode
+#   - the two right-most digits contain the actual opcode
+#   - any further digits contain the parameter mode of the parameters, reading digits from right-to-left
+#     and parameters in-order from left to right. Any unspecified digits default to 0 (position mode).
+#     NOTE: parameters for the target address of a write operation (e.g. the third parameter of the 'sum' or 'multiply' opcode)
+#           are never given in immediate mode.
```
thecodingcube77, line 31: I don't see where you implemented this. I think this is where your problem lies.

thecodingcube77 (commented):
You need to implement the two parameter modes.

DrKeks1010 (changes requested):
TheCodingCube77 is right, you are not supporting the two different input modes:

Currently your code throws an error at listCS[6], which at that time will have the code 1105.

This code should be interpreted as follows:

Take the last two digits. This will be the type of instruction that has to be done.
 -- Your command should execute a Jump-If-True opperation

The next digits will tell you how to read your inputs:
Continuing from the right, the next numbe in the code ('1105') is a 1. This indicates that the first input will be in immediate mode. This means, that the value is taken directly as input to the operation.
 -- The condition for the jump now becomes: if listCS[counter + 1] != 0
 -- If the input would be given in position mode (like with the code '1005' or '05'), you should test: if listCS[listCS[counter + 1]] != 0

The same pattern conitnues for the last digit:
This will again be a 1 (immediate mode). So the position to jump to will be directly given as paramater
 -- The jump should now execute and set: counter = listCS[temp + 2]
 -- Otherwise (with position mode: e.g. '0105', '105', '05' or '5') the jump should be executed as: counter = listCS[listCS[temp + 2]]


to get these input modes, I would advise you to use a structure like this:
temp = listCS[counter]
optCode = temp % 100 #This will return the last two digits of the code. '1105' becomes '05'
temp //= 100 #discard the now read optcode. '1105' becomes '11'
input_mode_1 = temp % 10 #returns the last digit. '11' becomes '1'
temp //= 10 #discard the last digit
input_mode_2 = temp % 10 #read the second input mode

this way you can then test for the input_mode and either use listCS[counter + 1] od listCS[listCS[counter + 1]] appropiatly

NOTE: You will have to do this fix for Add, Multiply, Output, Jump, Less Than and Equals.

PR #102, opened 2023-11-01
Lösungen Woche 4

s231365:
Please review:
solution to task: 16694270

### DIFF
--- a/Exercises/LauravanUffelt/VL 4 Lösungen.py
+++ b/Exercises/LauravanUffelt/VL 4 Lösungen.py
@@ -45,25 +45,30 @@ def comp_sim(intake):
     def element_value(mode, element):
         return intake[element] if mode == 0 else element
 
+    def elements(count, mode, intake):
+        element1 = element_value(mode[0], intake[count+1])
+        element2 = element_value(mode[1], intake[count+2])
+        return element1, element2
+
     def opcodes(count):
         operation = str(intake[count]).zfill(5)
         opcode = int(operation[-2:])
         mode = [int(digit) for digit in operation[:-2]][::-1]
-        element1 = element_value(mode[0], intake[count+1])
 
         match opcode:
+
             case 99:
                 exit(0)
 
             case 1:
-                element2 = element_value(mode[1], intake[count+2])
+                element1, element2 = elements(count, mode, intake)
                 sum = element1 + element2
                 adress = intake[count+3]
                 intake[adress] = sum
                 return count + 4
 
             case 2:
-                element2 = element_value(mode[1], intake[count+2])
+                element1, element2 = elements(count, mode, intake)
                 product = element1 * element2
                 adress = intake[count+3]
                 intake[adress] = product
@@ -81,7 +86,7 @@ def comp_sim(intake):
                 return count + 2
 
             case 5:  # jump-if-true: if the first parameter is non-zero, it sets the instruction pointer to the value from the second parameter. Otherwise, it does nothing.
-                element2 = element_value(mode[1], intake[count+2])
+                element1, element2 = elements(count, mode, intake)
                 if element1 != 0:
                     count = element2
                     return count
@@ -89,7 +94,7 @@ def comp_sim(intake):
                     return count + 3
 
             case 6:  # jump-if-false: if the first parameter is zero, it sets the instruction pointer to the value from the second parameter. Otherwise, it does nothing.
-                element2 = element_value(mode[1], intake[count+2])
+                element1, element2 = elements(count, mode, intake)
                 if element1 == 0:
                     count = element2
                     return count
@@ -97,7 +102,7 @@ def comp_sim(intake):
                     return count + 3
 
             case 7:  # less than: if the first parameter is less than the second parameter, it stores 1 in the position given by the third parameter. Otherwise, it stores 0.
-                element2 = element_value(mode[1], intake[count+2])
+                element1, element2 = elements(count, mode, intake)
                 adress = intake[count+3]
                 if element1 < element2:
 
@@ -107,7 +112,7 @@ def comp_sim(intake):
                     intake[adress] = 0
                     return count + 4
             case 8:  # equals: if the first parameters equal to the second parameter, it stores 1 in the position given by the third parameter. Otherwise, it stores 0.
-                element2 = element_value(mode[1], intake[count+2])
+                element1, element2 = elements(count, mode, intake)
                 adress = intake[count+3]
                 if element1 == element2:
                     intake[adress] = 1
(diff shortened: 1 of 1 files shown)


Exercises/LauravanUffelt/VL 4 Lösungen.py
```diff
@@ -0,0 +1,131 @@
...
+    def opcodes(count):
+        operation = str(intake[count]).zfill(5)
+        opcode = int(operation[-2:])
+        mode = [int(digit) for digit in operation[:-2]][::-1]
+        element1 = element_value(mode[0], intake[count+1])
+        element2 = element_value(mode[1], intake[count+2])
+
+        match opcode:
+            case 99:
+                return None
```
TrueEsel, line 57: you could also write something like "exit(0)" to immediately stop the program instead of returning none

Exercises/LauravanUffelt/VL 4 Lösungen.py
```diff
@@ -0,0 +1,131 @@
...
+# And here's some test cases:
+#   3,9,8,9,10,9,4,9,99,-1,8 -- test whether the input is equal to 8 (using position mode)
+#   3,3,1107,-1,8,3,4,3,99 -- test whether the input is less than 8 (using immediate mode)
+#   3,3,1105,-1,9,1101,0,0,12,4,12,99,1 -- test whether the input is 0 using jump instructions
+#
+# Finally, run you code for the following instructions; when asked for input provide the number '5'. The program should print a single number when executed.
+# Please take note of that number in your PR, so I don't need to run all the files myself :)
+def comp_sim(intake):
+    def element_value(mode, element):
+        return intake[element] if mode == 0 else element
```
TrueEsel, line 46: nice way to deal with the different read modes

Exercises/LauravanUffelt/VL 4 Lösungen.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def opcodes(count):
+        operation = str(intake[count]).zfill(5)
+        opcode = int(operation[-2:])
+        mode = [int(digit) for digit in operation[:-2]][::-1]
+        element1 = element_value(mode[0], intake[count+1])
+        element2 = element_value(mode[1], intake[count+2])
+
+        match opcode:
```
TrueEsel, line 55: im not sure if writing the whole logic for the operations into a match case statement is a good idea, even though its clearly working. You could use an dictionary with the required operations instead

Exercises/LauravanUffelt/VL 4 Lösungen.py
```diff
@@ -0,0 +1,131 @@
...
+#   3,3,1107,-1,8,3,4,3,99 -- test whether the input is less than 8 (using immediate mode)
+#   3,3,1105,-1,9,1101,0,0,12,4,12,99,1 -- test whether the input is 0 using jump instructions
+#
+# Finally, run you code for the following instructions; when asked for input provide the number '5'. The program should print a single number when executed.
+# Please take note of that number in your PR, so I don't need to run all the files myself :)
+def comp_sim(intake):
+    def element_value(mode, element):
+        return intake[element] if mode == 0 else element
+
+    def opcodes(count):
```
TrueEsel, line 48: In my opinion this name is a bit confusing since your whole logic is in this method. I'd expect that something with the name opcodes would only deal with processing an opcode, not executing all of the calculations.

Exercises/LauravanUffelt/VL 4 Lösungen.py
```diff
@@ -0,0 +1,131 @@
...
+                else:
+                    intake[adress] = 0
+                    return count + 4
+            case _:
+                print("ERROR")
+
+    count = 0
+    while count < len(intake)-2:
+        result = opcodes(count)
+        if isinstance(result, int):
```
TrueEsel, line 119: I dont understand what the logic from line 119 to 123 is supposed to achieve, since we dont need to return anything. I tried replacing these lines with "count = opcodes(count)" and it still worked.

TrueEsel (approved):
All in all the code is doing its job

Exercises/LauravanUffelt/VL 4 Lösungen.py
thecodingcube77: I think the idea is, that the opcode 99 returns `None`, which is no instance of `int`. But the else does nothing so the opcode 99 would get skipped I think. But the opcode 99 gets never reached because of the `len(intake) - 2` in line 117. The `- 2` should not be there. You probably added it because you got an `IndexError` in line 53. With some logic in place to stop the error from halting the execution the program reaches the opcode 99, your count gets set to `None` but you don't do anything in the else in line 122. This causes the program to end up in an endless loop. Not good. Line 122 should contain some code to properly halt the program.

thecodingcube77 (changes requested):
A proper halt to the program is not implemented at all. It only worked because the halt opcode is positioned at the second last index of the array passed to the function. If it would have been anywhere earlier you would have realized what the problem is.

Exercises/LauravanUffelt/VL 4 Lösungen.py
```diff
@@ -50,19 +50,20 @@ def opcodes(count):
...
         element1 = element_value(mode[0], intake[count+1])
-        element2 = element_value(mode[1], intake[count+2])
 
         match opcode:
             case 99:
-                return None
+                exit(0)
 
             case 1:
+                element2 = element_value(mode[1], intake[count+2])
```
thecodingcube77, line 59: Moving this line into the cases where it's needed creates much duplicate code. It could be placed in a function as suggested in my other comment. Duplicate code provokes bugs if you want to change something as you could forget to do the change in one place.

Exercises/LauravanUffelt/VL 4 Lösungen.py
```diff
@@ -114,13 +119,8 @@ def opcodes(count):
...
     count = 0
-    while count < len(intake)-2:
-        result = opcodes(count)
-        if isinstance(result, int):
-            count = result
-        else:
-            pass
-    return result
+    while count < len(intake):
+        count = opcodes(count)
```
thecodingcube77, line 123: This is way better.

Exercises/LauravanUffelt/VL 4 Lösungen.py
```diff
@@ -50,19 +50,20 @@ def opcodes(count):
         opcode = int(operation[-2:])
         mode = [int(digit) for digit in operation[:-2]][::-1]
```
thecodingcube77, line 51: You don't really need to determine the modes every time as not every opcode requires modes, could also be placed into a separate function.

thecodingcube77 (changes requested):
The code could still run into an error with a different array of commands with opcode 99 in the last place. This is due to the access to `intake[count + 1]` in line 52.
You also execute some code in every loop which actually is not needed for all the opcodes (the determination of (2)modes). 
For duplicate code the migration to a separate function is always a good idea.

thecodingcube77 (commented):
Everything is good now in my opinion. Requested changes successfully implemented.

thecodingcube77 (approved):
Well I should also approve

PR #104, opened 2023-11-02
Solution Week 4 [16694270]

CorvinAn97:


Exercises/Corvin Ansorge/exercises_04.py
```diff
@@ -0,0 +1,105 @@
...
+            i += 4
+        elif opcode == 2:
+            commands[commands[i+3]] = get_value(i+1, param_modes[0]) * get_value(i+2, param_modes[1])
+            i += 4
+        elif opcode == 3:
+            commands[commands[i+1]] = input_value
+            i += 2
+        elif opcode == 4:
+            output = get_value(i+1, param_modes[0])
+            i += 2
```
thecodingcube77, line 80: The output opcode is not supposed to determine the return value. You are supposed to print the read value exactly here.

Exercises/Corvin Ansorge/exercises_04.py
```diff
@@ -0,0 +1,105 @@
...
+
+        if opcode == 1:
+            commands[commands[i+3]] = get_value(i+1, param_modes[0]) + get_value(i+2, param_modes[1])
+            i += 4
+        elif opcode == 2:
+            commands[commands[i+3]] = get_value(i+1, param_modes[0]) * get_value(i+2, param_modes[1])
+            i += 4
+        elif opcode == 3:
+            commands[commands[i+1]] = input_value
+            i += 2
```
thecodingcube77, line 77: You should implement that you actually have to input a value if you get the corresponding opcode. It works in his case as you only need to input once, but strictly speaking is this incorrect.

Exercises/Corvin Ansorge/exercises_04.py
```diff
@@ -0,0 +1,105 @@
...
+
+def execute_program(commands, input_value):
+    i = 0
```
thecodingcube77, line 57: `i` should really not be used in such a large scope. A more meaningful name like `op_pointer` or something like that would be better.

thecodingcube77 (changes requested):
All in all good solution.

Exercises/Corvin Ansorge/exercises_04.py
CorvinAn97: fixed that

Exercises/Corvin Ansorge/exercises_04.py
```diff
@@ -53,53 +53,61 @@
...
-            commands[commands[i+3]] = get_value(i+1, param_modes[0]) * get_value(i+2, param_modes[1])
-            i += 4
+            commands[commands[op_pointer + 3]] = get_value(op_pointer + 1, param_modes[0]) * get_value(op_pointer + 2,
+                                                                                                       param_modes[1])
+            op_pointer += 4
         elif opcode == 3:
-            commands[commands[i+1]] = input_value
-            i += 2
+            commands[commands[op_pointer + 1]] = input_value
+            op_pointer += 2
```
thecodingcube77, line 81: Here is still no input taken from the user.

Exercises/Corvin Ansorge/exercises_04.py
```diff
@@ -53,53 +53,61 @@
...
         elif opcode == 4:
-            output = get_value(i+1, param_modes[0])
-            i += 2
+            output = get_value(op_pointer + 1, param_modes[0])
+            op_pointer += 2
```
thecodingcube77, line 84: The value you assign to output should be printed here.

atrejut:
@CorvinAn97 please try to get this old PR merged

thecodingcube77 (approved):

PR #106, opened 2023-11-02
Celina 4

CelinaRiegel:
The output was 16694270


Exercises/Celina/exercises_04.py
```diff
@@ -0,0 +1,189 @@
...
+                save_pos=first
+                commands[save_pos] = do_input()
+                i+=2
+                #print(save_pos,commands[save_pos])
+            case 4:
+                save_pos=first
+                do_output(save_pos,commands)
+                i+=2
+            case 5:
+                if jump_if_true(commands[first],commands[second]) != False:
```
LeMixer, line 167: Try to avoid calling the same function twice with the same parameters. Consider using the walrus operator or at least execute the function before the if statement and then check if `i` is true afterwards. Furthermore why not check if the condition is `True` and not if it is `not False`?

LeMixer (changes requested):
The code is working and yielding the correct result. You can implement my suggestions if you want, but some things like the misleading `first_pos` parameter should be changed. Also fix your folder structure. It seems like you accidentally created a new folder and uploaded your results there.

Exercises/Celina/exercises_04.py
```diff
@@ -0,0 +1,189 @@
...
+    if first_pos<second_pos:
+        return 1
+    else:
+        return 0
+    
+def equals(first_pos, second_pos):
+    if first_pos == second_pos:
+        return 1
+    else:
+        return 0
```
ClasherKasten, line 126: this whole function can be substituted by a simple
```python
int(first_pos == second_pos)
```

kuckuk3 (changes requested):

Exercises/Celina/exercises_04.py
```diff
@@ -0,0 +1,189 @@
...
+   
+
+def wierd_computer(commands):
```
ClasherKasten, line 129: typo in function name. probably should be `weird_computer`

ClasherKasten (changes requested):
Please do not hinder yourself from adding type annotations to your code, It'd be really useful.

Exercises/Celina/exercises_04.py
```diff
@@ -164,13 +142,13 @@ def wierd_computer(commands):
                 do_output(save_pos,commands)
                 i+=2
             case 5:
-                if jump_if_true(commands[first],commands[second]) != False:
-                    i=jump_if_true(commands[first],commands[second])
+                if jump_if(commands[first],commands[second]) == True:
```
ClasherKasten, line 145: `== True` is unnecessary in this context. Just remove it.

ClasherKasten (changes requested):
Old comments still apply.

atrejut:
@CelinaRiegel please try to get this old PR merged

LeMixer (approved):

PR #115, opened 2023-11-05
Jasmin05 (1639024365, 6260)

not-too-deranged:


Exercises/JasminSteidl/Übung5.py
```diff
@@ -0,0 +1,117 @@
...
+# The five preceding numbers are [95, 102, 117, 150, 182], and no possible sum of two of those
+# numbers adds to 127.
+#
+# You can find the ACTUAL input for this exercise under `data/input_sequence.txt`. For this
+# real input you should consider not only the 5 numbers, but the 25 numbers preceding.
+# Find the first number in this list which can not be expressed as a
+# sum of two numbers out of the 25 numbers before it.
+# Please make not of your result in the PR.
+
+file = open("input_sequence.txt", "r")
```
GeFritz, line 19: It is a good practice do close files after having worked with them. One good alternative is `with open(<path>) as <filename>:`. There errors are mitigated and the files do not need to be manually closed.

Exercises/JasminSteidl/Übung5.py
```diff
@@ -0,0 +1,117 @@
...
+                found = True
+                break
+
+        if not found:
+            return target_number
+
+        input_numbers.remove(number_sequence[index - 25])
+        input_numbers.add(target_number)
+
+    return target_number
```
GeFritz, line 41: This will be called if there is never an unsummed number. It would be simpler if the function returns None, as returning a summed number is not the task of the function (as interpreted by the functions name). If the return is left out, None is always returned.

Exercises/JasminSteidl/Übung5.py
```diff
@@ -0,0 +1,117 @@
...
+# As usual, please list the answer as part of the PR.
+
+
+def count_bags_inside(bag_input, bag_colour):
+    if bag_colour not in bag_input:
+        return 0
+    total_bags = 0
+
+    for inner_bag, quantity in bag_input[bag_colour].items():
+        total_bags += quantity + quantity * count_bags_inside(bag_input, inner_bag)
```
GeFritz, line 89: Recursion is a good solution to this problem.

GeFritz (changes requested):
All in all two clear and relatively simple solutions. The only thing that has to change is that the files are not closed, which could cause some drastic errors.

GeFritz (approved):

PR #119, opened 2023-11-05
Exercise 5 (1639024365)(6260)

RubelOS:
Task 1
1639024365 at index: 653

Task 2
6260


Exercises/JuliusDralle/exercises_05.py
```diff
@@ -0,0 +1,137 @@
...
+# You can find the ACTUAL input for this exercise under `data/input_sequence.txt`. For this
+# real input you should consider not only the 5 numbers, but the 25 numbers preceding.
+# Find the first number in this list which can not be expressed as a
+# sum of two numbers out of the 25 numbers before it.
+# Please make not of your result in the PR.
+
+def findNumber():
+    numbers = []
+
+    fp = open("data\input_sequence.txt", 'r')
```
thecodingcube77, line 22: You should consider using pathlib. I was unable to run this file on Linux due to the `\` in the Windows path.

Exercises/JuliusDralle/exercises_05.py
```diff
@@ -0,0 +1,137 @@
...
+    lines = fp.read().strip()
+
+    for currentLine in lines.split("\n"):
+        numbers.append(int(currentLine))
+
+    for iteratorThroughBigSteps in range(26,len(numbers),26):
+        for iteratorThroughSmallSteps in range(iteratorThroughBigSteps-26, iteratorThroughBigSteps):
+            for iteratorThroughSmallStepsTwo in range(iteratorThroughBigSteps-26, iteratorThroughBigSteps):
+                if numbers[iteratorThroughSmallSteps] + numbers[iteratorThroughSmallStepsTwo] == numbers[iteratorThroughBigSteps] and iteratorThroughSmallSteps != iteratorThroughSmallStepsTwo:
+                    return f"{numbers[iteratorThroughSmallSteps]} + {numbers[iteratorThroughSmallStepsTwo]} = {numbers[iteratorThroughBigSteps]}"
```
thecodingcube77, line 33: Your result of 21 is wrong. I think you understood the task in a wrong way. What you did is checking for the first pair of numbers that add up to the 27th number(also wrong, should be 26th) and end the execution there. But you should actually search for the first number that is not a sum of any 2 numbers in the 25 before it.
The step of 26 is also not correct as you now only check every 26th number but you should check every number(except for the first 25 as they don't have 25 numbers before them).
Next point. This code is very unreadable in my opinion. It took me quite some time to understand fully. Might be worth restructuring.
Ways to fix/improve:
    - Turn the 26's into 25's
    - Remove the step size
    - Change the if, that it now checks if no pair in your 25 numbers sums up to the given number

Exercises/JuliusDralle/exercises_05.py
```diff
@@ -0,0 +1,137 @@
...
+# Find the first number in this list which can not be expressed as a
+# sum of two numbers out of the 25 numbers before it.
+# Please make not of your result in the PR.
+
+def findNumber():
+    numbers = []
+
+    fp = open("data\input_sequence.txt", 'r')
+
+    lines = fp.read().strip()
```
thecodingcube77, line 24: You can use
```py
with open('filepath') as file:
    # do your stuff here with the file
```
to automatically close the file after the code block

Exercises/JuliusDralle/exercises_05.py
```diff
@@ -0,0 +1,137 @@
...
+            new_statement = i[2:-6]
+            statement_array.append(new_statement)
+        elif i[-3:] == "bag":
+            new_statement = i[2:-4]
+            statement_array.append(new_statement)
+        elif i[-4:] == "bag.":
+            new_statement = i[2:-5]
+            statement_array.append(new_statement)
+        else:
+            statement_array.append(str(i[3:]))
```
thecodingcube77, line 109: This can be very easy and readable if you use regex. It is quite easy in this simple case.

Exercises/JuliusDralle/exercises_05.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+statement_dict_array = [[],[]]  
+for iterator_lines in lines:
+    #print(iterator_lines.split(" contain "))
+    current_bag, statement = iterator_lines.split(" contain ")
+    current_bag = current_bag[:-5]
+    
+    
+    statement_array = []
```
thecodingcube77, line 91: `statement` is a confusing name in this context in my opinion.

Exercises/JuliusDralle/exercises_05.py
```diff
@@ -0,0 +1,137 @@
...
+#print(statement_dict_array)
+
+def countBags(bagType):
+    counter = 1
+    # look for recursion end
+    for i in range(0,len(statement_dict_array[0])):
+        if statement_dict_array[0][i] ==  bagType and statement_dict_array[1][i][0] == "n":
+            return 1
+    
+    for i in range(0,len(statement_dict_array[0])):
```
thecodingcube77, line 124: When starting at `0` the `range()` function actually does not need a start value.

thecodingcube77 (changes requested):
What you need to do is fixing the first task as your result is clearly wrong. The other comments are more of suggestions.

Exercises/JuliusDralle/exercises_05.py
RubelOS: The 26th Number is the first number which is being looked at

Exercises/JuliusDralle/exercises_05.py
```diff
@@ -0,0 +1,135 @@
...
+        lines = fp.read().strip()
+
+        for currentLine in lines.split("\n"):
+            numbers.append(int(currentLine))
+
+        isBuildable = True
+        for iteratorThroughBigSteps in range(25,len(numbers)):
+            isBuildable = False
+            for iteratorThroughSmallSteps in range(iteratorThroughBigSteps-26, iteratorThroughBigSteps):
+                for iteratorThroughSmallStepsTwo in range(iteratorThroughBigSteps-26, iteratorThroughBigSteps):
```
thecodingcube77, line 33: The -26 in the start points for the ranges should be 25, because you only want to check the 25 numbers before. You include the index -1 in the first iteration of the outer for loop because of this. The result stays the same but he logic is not 100% correct at the moment.

thecodingcube77 (approved):

atrejut:
Approved by @thecodingcube77 even though the solution to task 2 is different from your own solution? Only one of you can be right :D

RubelOS:
All problems focusing on functionality were solved. I did not implement a dictionary but still continue using my own nested list

Exercises/JuliusDralle/exercises_05.py
```diff
@@ -93,43 +93,44 @@ def findNumber():
...
+        for iterator_bagsinbags in range(0,len(statement_bags_array[0])):
+            if statement_bags_array[0][iterator_bagsinbags] ==  bag_type:
+                for multiplicator_of_bags in range(1,len(statement_bags_array[1][iterator_bagsinbags]),2):
+                    counter = counter + int(statement_bags_array[1][iterator_bagsinbags][multiplicator_of_bags-1]) * countBags(statement_bags_array[1][iterator_bagsinbags][multiplicator_of_bags])
 
         return counter  
 
-print(countBags("shiny gold"))
+#The -1 is subtracting the "shiny gold bag" out of the counted
+print(countBags("shiny gold")-1)
```
thecodingcube77, line 134: I consider this a really hacky solution. You now require the user to always subtract 1 from the result of the function, as it actually does not return the correct answer by itself. This change just resulted from the knowledge of the correct solution but would not have been made in a scenario where the answer is unknown.

thecodingcube77 (changes requested):
You don't need to change your data structure to a dictionary to make `countBags()` (still wrong format) work to return the right answer.

RubelOS:
I now use PathLib

PR #120, opened 2023-11-06
Bo Wichtendahl 05 (first task: 1639024365; second task: 6260)

thecodingcube77:


The-Linux-User (approved):
All seems good, nice solution to Task 2!

PR #121, opened 2023-11-07
Johannes Lingk Solution Week 5

joh-lin:
Answer 1: 1639024365
Answer 2: 6260


Reisii (approved):
Nice use of one liners, however, runtime will be a tiny bit too long but still a very good solution

PR #126, opened 2023-11-09
Robert Walz Woche05

LeMixer:
Task 1: 1639024365 at 653
Task 2: 6260


Exercises/RobertWalz/exercises_05/exercises_05.py
```diff
@@ -0,0 +1,124 @@
...
+# For the actual inputs, how many bags are inside your single shiny gold bag?
+# As usual, please list the answer as part of the PR.
+
+
+# Example usage:
+
+
+from bag import Bag
+
+def get_data():
```
paulseidel08, line 94: You shouldn't name two functions the same. It seems to create problems with running these.

paulseidel08 (changes requested):
Nice and short code!
But there are two problems:
1) You have two functions get_data(). Please rename at least one of those!
2) You have uploaded two versions of exercises_05.py thats a redundancy!

paulseidel08 (approved):
Everything fine now.

PR #131, opened 2023-11-09
H.Neu_ex.5

HenrietteNeumann:
exercise 5

solutions
part 1: 1639024365
part 2: 6260

### DIFF
--- a/exercise_5.py
+++ b/exercise_5.py
@@ -1,0 +1,87 @@
+# PART 1:
+# Here's a sequence of numbers:
+# [35, 20, 15, 25, 47, 40, 62, 55, 65, 95, 102, 117, 150, 182, 127, 219, 299, 277, 309, 576]
+# numbers in this list can in general be expressed as a sum of some pair of two numbers
+# in the five numbers preceding them.
+# For example, the sixth number (40) cam be expressed as 25 + 15
+# the seventh number (62) can be expressed as 47 + 15 etc.
+# 
+# The only exception to this rule for this example is the number 127.
+# The five preceding numbers are [95, 102, 117, 150, 182], and no possible sum of two of those
+# numbers adds to 127.
+#
+# You can find the ACTUAL input for this exercise under `data/input_sequence.txt`. For this
+# real input you should consider not only the 5 numbers, but the 25 numbers preceding.
+# Find the first number in this list which can not be expressed as a
+# sum of two numbers out of the 25 numbers before it.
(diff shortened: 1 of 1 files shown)


exercise_5.py
```diff
@@ -0,0 +1,87 @@
...
+    for line in file:
+        rule_parts = line.split()
+        outer_bag_color = f"{rule_parts[0]} {rule_parts[1]}"
+        inner_bags = [(int(rule_parts[i]), f"{rule_parts[i + 1]} {rule_parts[i + 2]}") for i in range(4, len(rule_parts), 4)]
+        rules[outer_bag_color] = inner_bags
+
+
+
+# Count the number of bags inside the target bag
+total_bags_inside = count_bags_inside(rules)
```
christianrietzsch, line 83: The method `count_bags_inside` needs a second argument: `bag_color`, which should be `"shiny gold"` for this task.

exercise_5.py
```diff
@@ -0,0 +1,87 @@
...
+    return sum(count * (1 + count_bags_inside(rules, color)) for count, color in rules.get(bag_color, []))
+
+
+# Read bag rules from file
+with open("data/input_bags.txt") as file:
+    rules = {}
+    for line in file:
+        rule_parts = line.split()
+        outer_bag_color = f"{rule_parts[0]} {rule_parts[1]}"
+        inner_bags = [(int(rule_parts[i]), f"{rule_parts[i + 1]} {rule_parts[i + 2]}") for i in range(4, len(rule_parts), 4)]
```
christianrietzsch, line 77: You need to check for bags that do not contain any other bags, like: 
`dotted black bags contain no other bags.` 
The program throws an error because you try to convert the element in` rule_parts[i]` to an integer, which is not possible for strings that do not contain any numbers:
You could add after `range()` : `if not rule_parts[i].startswith("no")`

exercise_5.py
```diff
@@ -0,0 +1,87 @@
...
+        current_number = sequence[i]
+        preamble = sequence[i - preamble_length:i]
+
+        if not any(current_number == x + y for j, x in enumerate(preamble) for y in preamble[j + 1:]):
+            return current_number
+
+
+# Sequence from file
+with open("data/input_sequence.txt") as file:
+    sequence = list(map(int, file.read().split()))
```
christianrietzsch, line 31: fast method to convert a list of strings to integers

christianrietzsch (changes requested):

PR #132, opened 2023-11-09
Tufgaben5 Aufgabe1: 1639024365 Aufgabe2: 6260

Reisii:


Exercises/TuNguyenXuan/Tufgaben5.py
```diff
@@ -0,0 +1,96 @@
...
+        return bagdict
+
+def count_of_bags(bag,bagdict):
+    count = 0
+    child = bagdict[bag].split(",")
+    for element in child:
+        numofelements = element[0]
+        if numofelements != "n":
+            count += int(numofelements)
+            for childcount in range(int(numofelements)):
```
fabischw, line 92: It looks like you're not using the childcount variable anywhere. You could use _ instead of childcount here (this is not mandatory though).

Exercises/TuNguyenXuan/Tufgaben5.py
```diff
@@ -0,0 +1,96 @@
...
+# You can find the ACTUAL input for this exercise under `data/input_sequence.txt`. For this
+# real input you should consider not only the 5 numbers, but the 25 numbers preceding.
+# Find the first number in this list which can not be expressed as a
+# sum of two numbers out of the 25 numbers before it.
+# Please make not of your result in the PR.
+
+
+
+
+with open('input_sequence.txt', 'r') as file:
```
fabischw, line 22: please use paths with the pathlib to access the files as the script you're providing does not have access to the files as they're located in a different folder.

Exercises/TuNguyenXuan/Tufgaben5.py
```diff
@@ -0,0 +1,96 @@
...
+
+def count_of_bags(bag,bagdict):
+    count = 0
+    child = bagdict[bag].split(",")
+    for element in child:
+        numofelements = element[0]
+        if numofelements != "n":
+            count += int(numofelements)
+            for childcount in range(int(numofelements)):
+                count_of_bags(element[1:], bagdict)
```
DrKeks1010, line 93: you should use count += count_of_bags(...) here, as you dont add the bags inside a subbag to the total count

Exercises/TuNguyenXuan/Tufgaben5.py
```diff
@@ -0,0 +1,96 @@
...
+    return count
+
+print(count_of_bags("shinygoldbag",get_bag_dict("input_bags.txt")))
```
fabischw, line 96: I do not know where you get the 6260 for the second task. Running your code 'as-is' this print statement prints out 7 (again, only after copying the necessary files from ../../data/).

Exercises/TuNguyenXuan/Tufgaben5.py
DrKeks1010, line 92: you might consider using
for _ in range(int(numodelements)):
here, so you are discarding the variable, as you are not using it

DrKeks1010 (changes requested):
Great solution. One small mistake.

fabischw (changes requested):
Your code does not yield the correct results 'as-is'. Please adjust the code to make it work without having to copy files and adjust the second task to print out the correct result.

Exercises/TuNguyenXuan/Tufgaben5.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+
+with open('input_sequence.txt', 'r') as file:
+    list1 = []
+    for line in file:
+        list1.append(int(line))
+
+def unsummable(seq):
+    input = set(seq[:25])
```
ClasherKasten, line 28: Don't use a builtin function name as variable name.

ClasherKasten (changes requested):
Type annotations are missing.

PR #140, opened 2023-11-13
Aufgabe06 Jasmin 1778099, 1623571

not-too-deranged:
Ich hab da völlig falsche Werte raus, weiß aber nicht wieso, wenn also jemand sich das anschauen könnte und Veränderungen vorschlagen könnte, wäre ich sehr dankbar.

### DIFF
--- a/Exercises/JasminSteidl/Übung6.py
+++ b/Exercises/JasminSteidl/Übung6.py
@@ -97,8 +97,12 @@
 # 
 # You can find the /actual/ input to both parts in data/terminal_record.txt
 ###############################################################################
-input = open('terminal_record.txt').read().splitlines()
+with open('terminal_record.txt') as f:
+    task_input = f.read().splitlines()
 
+MAX_SIZE_FOR_ADDING = 100000
+MAX_SIZE_DIR = 70000000
+NEEDED_SPACE = 30000000
 
 class Dir:
     def __init__(self, name, parent):
@@ -132,8 +136,7 @@ def calc_dir_sizes(dir):
 
 
 calc_dir_sizes(root)
-print("The Sum of all the direcotries under 100000:", sum([size for size in dir_sizes if size < 100000]))
+print("The Sum of all the direcotries under 100000:", sum([size for size in dir_sizes if size < MAX_SIZE_FOR_ADDING]))
 
-
-needed_space = 30000000 - (70000000 - dir_sizes[-1])
-print("Deleted file size:", min([size for size in dir_sizes if size >= needed_space]))
+freed_space = NEEDED_SPACE - (MAX_SIZE_DIR - dir_sizes[-1])
+print("Deleted file size:", min([size for size in dir_sizes if size >= freed_space]))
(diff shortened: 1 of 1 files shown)


Exercises/JasminSteidl/Übung6.py
```diff
@@ -0,0 +1,160 @@
...
+    with open(file_path, 'r') as file:
+        for line in file:
+            line = line.strip()
+            match = re.match(r'(dir\s\S+)|(\d+\s\S+)', line)
+            if match:
+                item = match.group(0)
+
+                if item.startswith('dir'):
+                    current_directory = current_directory + "/" + item.split()[1]
+                    stack.append((current_directory, 0))
```
thecodingcube77, line 117: With this code you go into a directory even though the `cd` command did not occur. The problem is as follows:
You have a directory `a` in that you execute `ls`. This shows you all the files and directories present in there. Files don't matter for the example right now.
`$ cd /` change to root directory
`$ cd a` go into directory `a`
`$ ls` list all files and directories present in directory `/a`
`dir b` directory called `b` at location `/a`
`dir c` directory called `b` at location `/a`
This should result in following paths of the directories:
-`/`
-`/a`
-`/a/b`
-`/a/c`
But in your case you get:
-`/`
-`/a`
-`/a/b`
-`/a/b/c`
See how the last path is different(and thus incorrect)? This is due to the directory change after getting a directory listed instead of a directory change after the `cd` command

Exercises/JasminSteidl/Übung6.py
```diff
@@ -0,0 +1,160 @@
...
+
+    with open(file_path, 'r') as file:
+        for line in file:
+            line = line.strip()
+            match = re.match(r'(dir\s\S+)|(\d+\s\S+)', line)
+            if match:
+                item = match.group(0)
+
+                if item.startswith('dir'):
+                    current_directory = current_directory + "/" + item.split()[1]
```
thecodingcube77, line 116: When the current directory is `/`, the resulting directory path is starting with a `//`.

thecodingcube77 (changes requested):
You completely ignored the commands `ls` and mainly important `cd`, which is where you should actually change the directory.

Exercises/JasminSteidl/Übung6.py
```diff
@@ -0,0 +1,142 @@
...
+root = Dir('/', None)
+current_dir = root
+
+for line in input[1:]:
+    line_parts = line.split(' ')
+    if line.startswith('dir'):
+        current_dir.subdirs[line_parts[1]] = Dir(line_parts[1], current_dir)
+    if line.startswith('$ cd'):
+        current_dir = current_dir.parent if line_parts[2] == '..' else current_dir.subdirs[line_parts[2]]
+    if line_parts[0].isnumeric():
```
thecodingcube77, line 124: The `if`'s could be changed to `elif`'s

thecodingcube77 (approved):
Solid solution. Very clean.

PR #142, opened 2023-11-15
Gerrit Fritz Aufgabe 6 (1778099, 1623571)

GeFritz:


Exercises/GerritFritz/exercises_06.py
```diff
@@ -0,0 +1,70 @@
+#Gerrit Fritz
+#15.11.2023
+
+from __future__ import annotations
+
+class Directory():
```
ClasherKasten, line 6: Python 3 doesn't require the parentheses after the class name.

Exercises/GerritFritz/exercises_06.py
```diff
@@ -0,0 +1,70 @@
...
+            
+    def update_size(self, size: int) -> None:
+        self.size += size
+        if self.parent != None: self.parent.update_size(size) 
+    
+    def get_dir(self, folder: str) -> Directory:
+        if folder in self.folders.keys(): return self.folders[folder]
+        elif folder == "..": return self.parent
+        elif folder == "/": return self.root
+        else: raise RuntimeError(f"No such dir: {folder}")
```
ClasherKasten, line 30: Please refrain from writing blocks like that. It's not very pythonic to do it, especially not to such an extent.

Exercises/GerritFritz/exercises_06.py
```diff
@@ -0,0 +1,70 @@
...
+
+def get_mindir(size_dict:dict, max:int, free:int)->int: 
+    return min([val for val in size_dict.values() if val >= size_dict["/"]+free-max])
+
+    
+
+root = build_filesys(get_record("data//terminal_record.txt"))
+size_dict = root.get_dict()
+print(get_sum(size_dict))
+print(get_mindir(size_dict,70000000, 30000000))
```
ClasherKasten, line 70: No magic values!

ClasherKasten (changes requested):
Almost good type annotations but types inheriting from `Generic` should have concrete types.

Exercises/GerritFritz/exercises_06.py
GeFritz: Could you please elaborate?

Exercises/GerritFritz/exercises_06.py
GeFritz: What exactly is the problem here? Is not merely a preference?

Exercises/GerritFritz/exercises_06.py
ClasherKasten: Someone without access to the assignment text shouldn't have to guess why exactly these values are chosen.

GeFritz:
Thank you for the insights. The change requests are helpful, but normally the classmates have to review and get grades. The problem now is that the others have to review code that already was reviewed. I never saw a pull request from you, so I assume that you aren't in our course?

Exercises/GerritFritz/exercises_06.py
ClasherKasten: You probably won't stumble a lot about productive code like that. Which in the end means for you to adapt to the new (way more common and standardized) pattern.

See [PEP8](https://peps.python.org/pep-0008/#other-recommendations)
> Compound statements (multiple statements on the same line) are generally discouraged

ClasherKasten:
> The problem now is that the others have to review code that already was reviewed.

Not really a problem. Your course can have special standards which could invalidate/complement the review.

LeMixer:
> > The problem now is that the others have to review code that already was reviewed.
> 
> Not really a problem. Your course can have special standards which could invalidate/complement the review.
> 
> 

Yes, it is a problem. The weekly reviews a part of our graded weekly tasks. This repo is meant to be a platform for our class to practice writing and reviewing Python code. Therefore it's not your business with whatever argumentation to leave reviews potentially disallowing us from writing a thorough review on this pull request. So please refrain from writing reviews in this repo in the future!

GeFritz:
> Almost good type annotations but types inheriting from `Generic` should have concrete types.

what exactly should i change?

Reisii (approved):
Nice work. Code works as intended and nice use of classes. Imo type annotations aren't necessary, so no reason for it to be a reason to request changes.

PR #143, opened 2023-11-15
Solution to Task 1 and 2 (1778099 | 1623571)

RubelOS:


Exercises/JuliusDralle/exercises_06.py
```diff
@@ -0,0 +1,288 @@
...
+#print(input_lines)
+
+current_directory = None
+root_directory = None
+
+######################################################################################
+
+def command_changedir(target):
+    global current_directory
+    global root_directory
```
LeMixer, line 180: As mentioned previously refrain from using `global` if not absolutely necessary.

Exercises/JuliusDralle/exercises_06.py
```diff
@@ -0,0 +1,288 @@
...
+    elif target == "..":
+        current_directory = current_directory.get_parent()
+        return 0
+    elif target == "/":
+        current_directory = root_directory
+        return 0
+    else:
+        for searching_dir in current_directory:
+            if searching_dir.get_name() == target:
+                current_directory = searching_dir
```
LeMixer, line 194: This works, but can be solved more elegantly and efficiently with `next()` for example.

Exercises/JuliusDralle/exercises_06.py
```diff
@@ -0,0 +1,288 @@
...
+                return 0
+    raise ValueError("Directory not Found or not created, try creating one first!")
```
LeMixer, line 196: Think about what you are doing here. Would you really want your file system to crash the instance you `cd` into a non existing folder? If so leave it like that, if not don't raise the error or handle it.

Exercises/JuliusDralle/exercises_06.py
```diff
@@ -0,0 +1,288 @@
...
+put_all_dir_sizes_in_list(root_directory)
+list_of_directories.sort()
+
+
+needed_size = 0
+needed_size += root_directory.get_calculated_size()
+needed_size = 70000000 - needed_size
+needed_size = 30000000 - needed_size
+
+print(search_for_bigenough_dir(needed_size))
```
LeMixer, line 286: The task is debatable, but personally I read the task in the way that you have to return the dir as well not only it's size.

LeMixer (changes requested):
I would definitely change the `getters` and maybe some other stuff, if you have the time for it.

Exercises/JuliusDralle/exercises_06.py
fabischw: Well, we're getting a terminal log. So if there is no error message shown in the terminal we do not need to handle it. The code we're writing simply prints us a file structure. It doesn't represent / replace a file-system. The handling of this case is done by the actual file system and would result in an error message being logged as well. Handling an error message is near-impossible as we do not know what text the error message would contain. The critique here could be that his code is not able to handle anything but cd, ls dirs and files and doesn't handle potential error messages that might be logged.

fabischw: This also means that the error message is kind of misleading as the it should state that your code was not able to read the terminal log properly. The code has nothing to do with the file structure other than the fact that we want to print the structure of it. We do not have access to the actual filesystem.

RubelOS:
@LeMixer do u still request changes?

LeMixer (approved):
Looks good to me now.

PR #144, opened 2023-11-15
Robert Walz Woche06

LeMixer:
Task 1: 1778099
Task 2: ('/pcqjnl/lrrl/nwjggvr/bwmglvmt'), 1623571)

### DIFF
--- a/Exercises/RobertWalz/exercises_06/main.py
+++ b/Exercises/RobertWalz/exercises_06/main.py
@@ -1,4 +1,6 @@
 from enum import Enum
+from pathlib import Path
+import bisect
 
 from classes import Directory, File, FileSystem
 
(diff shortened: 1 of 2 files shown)


The-Linux-User:
Quite a lot of code, especially with the two separate files but seems to work all in all.

Exercises/RobertWalz/exercises_06/main.py
```diff
@@ -0,0 +1,94 @@
...
+    return count
+
+
+data = get_data()
+file_system = FileSystem(70000000)
+evaluateTerminalRecord(data=data, fs=file_system)
+dirs: dict[Path, int] = {}
+
+
+get_dir_sizes(file_system.current_dir.parent, dirs)
```
fabischw, line 82: Does this line server any purpose ? There seems to be a return statement in the function called, however the returned value is not safed in a variable.

Exercises/RobertWalz/exercises_06/main.py
```diff
@@ -0,0 +1,94 @@
...
+            sizes_of_dirs[content.path] = content.size
+            dir.size += get_dir_sizes(content, sizes_of_dirs=sizes_of_dirs)
+            sizes_of_dirs[dir.path] = dir.size
+
+    return dir.size
+
+
+def get_size_of_dirs_above_size(size: int, directories: dict[Path, int]):
+    values = [value for value in directories.values() if value <= size]
+    count = sum(values)
```
fabischw, line 72: Why not sum them immediatly instead of first creating a list and then get the sum ?

fabischw (approved):
Looks good in general, consider adding some more comments in the future and take a look at the comments I added.

Exercises/RobertWalz/exercises_06/main.py
LeMixer: Apparently it does. I am guessing it fills the size value for each dict. The return is there for a recursion.

Exercises/RobertWalz/exercises_06/main.py
LeMixer: Yeah, that's true

PR #147, opened 2023-11-15
Johannes Lingk Week 6 (1778099, 1623571)

joh-lin:


FritzRa (approved):
Great use of type annotations, complex and detailed class system to display the file system. Naming of variables and functions well done. Commendable usage of docstrings. All in all readable and as far as i can see without flaws.

PR #148, opened 2023-11-15
Laura Lösungen Woche 6 (1623571,1778099)

s231365:
task 1: 1623571
task 2: 1778099


fabischw (changes requested):
Code looks good, however you should use pathlib / os to read the file as your current code depends on the txt file being in the same directory as the python file which is NOT the case. Instead you should use something like this :
```py
here = pathlib.Path(__file__).parent
exercise_dir = here.parent

with open(data_dir / "terminal_record.txt","r") as file:
    data = file.readlines()

```

You should also remove whitespaces in the file name as they are often annoying to deal with.

PR #151, opened 2023-11-16
Solution Exercise_06

CorvinAn97:
Part 1: 1778099
Part 2: 1623571

### DIFF
--- a/Exercises/Corvin Ansorge/exercises_06.py
+++ b/Exercises/Corvin Ansorge/exercises_06.py
@@ -123,7 +123,7 @@ class Dir:
 root = Dir("/", None)
 curr_dir: Dir = root
 
-cmds = [e.strip() for e in open("/Users/corvinansorge/PycharmProjects/ppp-2023/data/terminal_record.txt") if not e.startswith("$ ls")]
+cmds = [e.strip() for e in open("./../../data/terminal_record.txt") if not e.startswith("$ ls")]
 for line in cmds[1:]:
     if line.startswith("$ cd"):
         curr_dir = curr_dir[line.split(" ")[2]]
--- a/Exercises/Corvin Ansorge/exercises_06.py
+++ b/Exercises/Corvin Ansorge/exercises_06.py
@@ -1,0 +1,147 @@
+# You get a recording of a terminal session moving around a file system.
+# The filesystem consists of a tree of files (plain data) and directories (which can contain other directories or files). 
+# The outermost directory is called /. You can navigate around the filesystem, moving into or out of directories
+# and listing the contents of the directory you're currently in.
+#
+# Within the terminal output, lines that begin with $ are commands you executed, very much like some modern computers:
+#
+# cd means change directory. This changes which directory is the current directory, but the specific result depends on the argument:
+# cd x moves in one level: it looks in the current directory for the directory named x and makes it the current directory.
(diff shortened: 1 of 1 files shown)


Exercises/Corvin Ansorge/exercises_06.py
```diff
@@ -0,0 +1,147 @@
...
+# Mention the total size of the deleted directory in the PR.
+# 
+# You can find the /actual/ input to both parts in data/terminal_record.txt
+###############################################################################
+
+from typing import Union
+
+
+class Dir:
+    def __init__(self, name: str, parent: Union["Dir", None]):
```
LeMixer, line 106: Instead of union you could use `Dir | None` instead which recommend by the python docs ([here](https://docs.python.org/3/library/typing.html))

Exercises/Corvin Ansorge/exercises_06.py
```diff
@@ -0,0 +1,147 @@
...
+        self.name = name
+        self.parent = parent
+        self.children: dict[str, "Dir"] = {}
```
LeMixer, line 109: This annotation is also considered to be somewhat legacy and should in general be done as follows: 

> In general, if something returns self, [...], you should use Self as the return annotation. [here](https://docs.python.org/3/library/typing.html)

This is true for all kind of type hints of a class inside itself, since Python 3.11. You can find it at `typing.Self`

Exercises/Corvin Ansorge/exercises_06.py
```diff
@@ -0,0 +1,147 @@
...
+root = Dir("/", None)
+curr_dir: Dir = root
+
+cmds = [e.strip() for e in open("/Users/corvinansorge/PycharmProjects/ppp-2023/data/terminal_record.txt") if not e.startswith("$ ls")]
+for line in cmds[1:]:
+    if line.startswith("$ cd"):
+        curr_dir = curr_dir[line.split(" ")[2]]
+        continue
+    if line.startswith("dir"):
+        d = Dir(line.split(" ")[1], curr_dir)
```
LeMixer, line 132: You should generally be using more descriptive names for variable, as the time save from writing a couple letters less is generally not worth it regarding readability. In this case it's not a huge issue, but it's definitely something to keep in mind.

Exercises/Corvin Ansorge/exercises_06.py
```diff
@@ -0,0 +1,147 @@
...
+        return sum([e for e in self.files]) + sum([len(e) for e in self.children.values()])
+
+    def __getitem__(self, item):
+        return self.parent if item == ".." else self.children[item]
+
+
+root = Dir("/", None)
+curr_dir: Dir = root
+
+cmds = [e.strip() for e in open("/Users/corvinansorge/PycharmProjects/ppp-2023/data/terminal_record.txt") if not e.startswith("$ ls")]
```
LeMixer, line 126: You should definitely change the absolute path to be at least somewhat of a relative path. This way users/reviewers have not chance running your code, if they'd change their cwd.

LeMixer (changes requested):
Generally a pretty clean code, but stay away from using the now old type hints I've mentioned, but more importantly give users/reviewers a chance of running your code.

LeMixer (approved):
Looks good now, feel free to merge

PR #153, opened 2023-11-16
Fabian woche6 (1778099 | 1623571)

fabischw:
Not entirely finished, still debugging, can't seem to find the bug. Offering a pack of 'Lebkuchen' to the first person to find the bug (if I don't find it first that is). Hint: it must be in the code that creates the dir_sizes_dict .


TomKnoden:
The issue in the code lies in the task1_get_max_dir_sizes function, if I'm getting it right. The problem is that the loop is iterating over the keys of the dictionary (dir_sizes_dict), but it should iterate over the values instead.

fabischw:
The loop is iterating over the keys and then I'm getting the value using that key, I just realised I could instead iterate over the values since I only needthe key for a section of the code that I removed so it's just dead code. It is not the issue unfortunately as I get the value eitherway. As I pointed out in my initial comment, I conducted testing that indicated that the 'dir_sizes_dict' is already incorrect but I still couldn't figure out why that is. Been working on the issue and conducted more tests whose results support the theory that something with the code block starting in line 161 that goes back and adjusts the parent directories seems to be incorrect.

Exercises/FabianSchwarz/exercises_06.py
```diff
@@ -0,0 +1,296 @@
...
+                    dir_sizes_dict[path_str] = curr_dir_size + int(file_size_str)#TODO: remove second type cast and cast into a int just once
+                    
+                    #go back and add the current file size to the parent directories
+                    if path_str.count("/") > 1:
+                        #get recursive path array
+                        progressive_path = [path_str[:i].rstrip('/') for i in range(1, len(path_str) + 1) if path_str[i - 1] == '/' or i == len(path_str)][1:]
+                        for path in progressive_path:
+                            curr_dir_size = dir_sizes_dict.get(path)
+                            curr_dir_size = 0 if not curr_dir_size else curr_dir_size
+                            dir_sizes_dict[path] = curr_dir_size + int(file_size_str)
```
GeFritz, line 168: In this loop, the parent directory sizes are updated. This iterates from the topmost directory to the current directory. The mistake is, that the current directory already has a size assigned to it (ll. 156-159).

GeFritz, line 168: A lot of variables and deep nesting. I would "outsource" parts of the code to specific functions for readability. Its good that the variables had fitting names. The overwhelming number of variables was not making bug hunting any easier.

GeFritz (changes requested):
All in all a very intriguing solution. The bug has been mentioned in the comments. I would personally adjust the nesting layers by outsourcing to "helper" functions. This would assist readability. Nice choice of variable names, this actually helped quite a bit with debugging. I assume the other file in this pull request will be removed.

Exercises/FabianSchwarz/exercises_06.py
```diff
@@ -0,0 +1,296 @@
...
+                    file_ending = file_name[file_name.find("."):]
+
+
+                    #save file ending
+                    if not file_ending.startswith("."):
+                        file_ending = ""
+                    else:
+                        file_ending = file_ending.replace(".","")
+
+                    file_name += file_ending
```
GeFritz: Forgot to mention, but this is seems to have no purpose whatsoever.

Exercises/FabianSchwarz/exercises_06.py
fabischw: Thanks for pointing that out, this was the bug that broke my code.

fabischw:
Alright, thanks for pointing out the bug, code should be working now, going to do dead code elemination and refactoring at a later date as the minimum requirements for the code should be fullfilled at the current stage of the pull request.

GeFritz (approved):
Nice and quick fix. Nice work.

PR #154, opened 2023-11-16
Week 6 Utku Acikel

UtkuDHBW:
Not finished, still trying to fix error in directory_sizes & part 2!


s231365 (commented):
Your solution for task 1 is working and is a good solution. Please make comments, so it is easier for the reviewer to understand your code.

PR #155, opened 2023-11-16
Aufgaben zu Woche 06 (1778099;1623571)

TomKnoden:


Exercises/Tom_Knoden/Aufgaben_Woche_06.py
```diff
@@ -0,0 +1,97 @@
...
+
+def find_directory_to_delete(directory_sizes, required_space):
+    for directory, size in sorted(directory_sizes.items(), key=lambda x: x[1]):
+        if size >= required_space:
+            return directory, size
+
+    return None, 0
+
+def main():
+    console_session_file = 'data/terminal_record.txt'
```
paulseidel08, line 81: Please use PathLib for importing the input file, so your code works for reviewers. You can just copy the file-importation from someone else, and edit the file name there.

paulseidel08 (changes requested):
Your code seems functional and efficient to me. Good job!
However, I have to request changes, because your file import doesn't work for everyone. Please use PathLib for that.

TomKnoden:
I accidentaly resolved the conversation regarding your last change request...I hope that does not cause any further problems.
Nonetheless, I think I've also already adressed your change request regarding the use of PathLib in my code.
Is the PR ready to merge, or are there any unsolved problems left?

paulseidel08 (approved):
Good idea, to just ask the User for the file-path just via input()! 
Everything seems fine now.

PR #156, opened 2023-11-16
Celina 6

CelinaRiegel:
Part 1: 1778099
Part 2: 1623571


Exercises/Celina/exercises_06.py
```diff
@@ -0,0 +1,165 @@
...
+
+file_path = []
+
+directory_dic={}
+
+
+console_session = open('terminal_record.txt', 'r')
+input = []
+for line in console_session:
+    input.append(line)
```
TrueEsel, line 110: you should always close a file that your opening. Consider doing
```suggestion
with open('terminal_record.txt', 'r') as console_session:
    input = []
    for line in console_session:
        input.append(line)
```
that way the file will automatically be closed
Also having to copy the terminal_record.txt into the working directory is not a nice solution, although its functioning. you could use pathlib to use a relative path.

Exercises/Celina/exercises_06.py
```diff
@@ -0,0 +1,165 @@
...
+                file_path.append(direc)
+        else:
+            output = line.split()
+            size = int(output[0])
+            path= ""
+            for elements in file_path:
+                path += "/"+elements   
+                if path in directory_dic:
+                    new_size = directory_dic[path] + size
+                    directory_dic[path] = new_size
```
TrueEsel, line 134: this can be simplified to `directory_dic[path] += size` improving readability

Exercises/Celina/exercises_06.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+
+def size_of_directorys(input):
+    #print(input)
+    for line in input:
+        if line.startswith("$ ls") or line.startswith("dir"):
+            continue
+        if line.startswith("$ cd"):
+            command = line.split()
```
TrueEsel, line 120: i did not consider to simply use split to solve this problem. Nice solution

Exercises/Celina/exercises_06.py
```diff
@@ -0,0 +1,165 @@
...
+            output = line.split()
+            size = int(output[0])
+            path= ""
+            for elements in file_path:
+                path += "/"+elements   
+                if path in directory_dic:
+                    new_size = directory_dic[path] + size
+                    directory_dic[path] = new_size
+                else:
+                    directory_dic[path] = size
```
TrueEsel, line 136: Easy and readable way to calculate the size of every directory. Great solution!

Exercises/Celina/exercises_06.py
```diff
@@ -0,0 +1,165 @@
...
+    if size_to_delete < 0:
+        print("nothing to delete")
+    else:
+        direc_to_delete=[]
+        for element in directory_dic:
+            if directory_dic[element]>size_to_delete:
+                direc = (element,directory_dic[element])
+                direc_to_delete.append(direc)
+        #print(direc_to_delete)
+        direc_to_delete.sort(key=lambda x: x[1])
```
TrueEsel, line 158: nice way to get the directory that should be deleted

Exercises/Celina/exercises_06.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+console_session = open('terminal_record.txt', 'r')
+input = []
+for line in console_session:
+    input.append(line)
+
+
+
+def size_of_directorys(input):
```
TrueEsel, line 114: while this function is generating a dictionary listing the size of each directory, its technically not returning this dictionary. Since you have initialized the dictionary on a global scale your code is still working, however i believe initializing the dictionary in this function and giving it to your other functions as a parameter would be a nicer solution.

TrueEsel (changes requested):
Great solution overall. However, you technically missed a small part of the assignment and should always close a file your opening.

TrueEsel (approved):
looks good now :)

PR #161, opened 2023-11-18
Gerrit Fritz Aufgabe 7 (17159)

GeFritz:
Highscore: 17159
Using Pyglet, older commits with Tutle/Tkinter
breakout

https://github.com/dhbw-ma-ppp/ppp-2023/assets/140959846/7a287186-ee4d-4463-83d0-a91191dd3c09


Exercises/GerritFritz/computer.py
```diff
@@ -0,0 +1,98 @@
...
+    def mode_to_index(self):
+        self.positions  = []
+        for i in range(0,len(self.mode)):
+            if len(self.commands)>self.memory_pointer+i+1:
+                local_mode = self.mode[i]
+                match local_mode:
+                    case "0": self.positions.append(self.commands[self.memory_pointer+i+1])
+                    case "1": self.positions.append(self.memory_pointer+i+1)
+                    case "2": self.positions.append(self.offset+self.commands[self.memory_pointer+i+1])
+                    case _: raise KeyError(f"Wrong mode {self.mode}")
```
fabischw, line 74: This is possible but not very pythonic. Using a new line for the code makes it more readable and does no harm.

Exercises/GerritFritz/computer.py
```diff
@@ -0,0 +1,98 @@
...
+            else:
+                raise IndexError(f"out of range")
```
fabischw, line 76: Raising a index error with that description is kind of pointless since the code would just raise an IndexError on itself if you didn't sort out that case. You could at least provide more information here if you want to raise an error yourself. Examples of informaiton you could print here are the values of self.commands and self.memory_pointer.

Exercises/GerritFritz/exercises_07.py
```diff
@@ -0,0 +1,118 @@
+from computer import Computer
+import pyglet
+from pyglet import shapes
```
fabischw, line 3: Why do this ? You use pyglet.ClassMethod for everything except the shapes. This is merely confusing. Either import the entire module and use pyglet.ClassMethod or import only the Classes / Methods you need and call the method / class itself.

fabischw (changes requested):
Please add a requirements.txt so users / reviewers can easily view and install your dependencies before seeing the error message. The path is also an issue. While your code is very readable, comments would make a great addition. The running program itself looks good.

Requested changes:
- change path to file location dependent path so it works for every cwd
- add requirements.txt

Optional / suggested changes:
- comments
- pyglet import
- error raising
- shift match case code block from same line as the case to line below

fabischw (approved):
LGTM!

RubelOS:
Did u speed up the video or is your program that efficient?

GeFritz:
> Did u speed up the video or is your program that efficient?

It should be about real time. Speed is high due to the library `pyglet`.

PR #162, opened 2023-11-19
Jasmin_Steidl_Jan_Merz_Exercise_07 (Score: 17159)

TrueEsel:
This task was pair programmed by Jasmin Steidl and Jan Merz.

> While the first part of the task is working fine, the second task is not working how its supposed to and we cant find the Error.
> The way the code is written right now the program works for two frames and is then exited because of opcode 99. 
> In these two frames the ball is moving towards the lower right part of the window and the paddle is moving in the same direction.
> If the input_getter function is returning zero from the start (meaning the paddle is not moving at all) the ball is going up and breaking a block first, instead of going straight down.
> We couldnt find out why this is the case.

Update:
Fixed several bugs in the program, including using the wrong coordinates to calculate the coordinate of the ball. The program is fully functioning now

Picture required by task 1:
start

### DIFF
--- a/Exercises/JanMerz/exercises_07.py
+++ b/Exercises/JanMerz/exercises_07.py
@@ -1,0 +1,255 @@
+import numpy as np
+import matplotlib.pyplot as plt
+# to solve todays exercise you will need a fully functional int-computer
+# a fully functional int-computer has some additional features compared to the 
+# last one you implemented.
+# you can either use my implementation of an int-computer (see AtrejuTauschinsky/int_computer.py)
+# or you can extend your own int-computer with the necessary features. If you want to extend your own
+# the features you need to implement will be described at the bottom.
+#
+#
+# We will run 'breakout' -- the arcade game -- on our simulated computer. 
+# (https://en.wikipedia.org/wiki/Breakout_(video_game))
+# The code for the computer will be provided under data/breakout_commands.txt
+# the code will produce outputs in triplets. every triplet that is output
+# specifies (x-position, y-position, tile_type).
+# tiles can be of the following types:
+# 0: empty tile
+# 1: wall. walls are indestructible
+# 2: block. blocks can be destroyed by the ball
+# 3: paddle. the paddle is indestructible
+# 4: ball. the ball moves diagonally and bounces off objects
+# 
+# EXAMPLE:
+# a sequence of output values like 1, 2, 3, 6, 5, 4 would
+#  - draw a paddle (type 3) at x=1, y=2
+#  - draw the ball (type 4) at x=6, y=5
+#
+#
+# PART 1:
+# run the game until it exits. Analyse the output produced during the run, and create
+# a visual representation (matplotlib or ascii-art are possibilities here...) of the screen display.
+# mark the different tile types as different colors or symbols. Upload the picture with your PR.
+#
+# PART 2:
+# The game didn't actually run in part 1, it just drew a single static screen.
+# Change the first instruction of the commands from 1 to 2. Now the game will actually run.
+# when the game actually runs you need to provide inputs to steer the paddle. whenever the computer
+# requests you to provide an input, you can chose to provide
+# -  0: the paddle remains in position
+# - -1: move the paddle to the left
+# - +1: move the paddle to the right
+#
+# the game also outputs a score. when an output triplet is in position (-1, 0) the third value of
+# the triplet is not a tile type, but your current score.
+# You need to beat the game by breaking all tiles without ever letting the ball cross the bottom 
+# edge of the screen. What is your high-score at the end of the game? provide the score as part of your PR.
+#
+# BONUS: (no extra points, just for fun)
+# make a movie of playing the game :)
+#
+#
+# COMPLETE INT COMPUTER
+# This is only relevant if you decide to extend your own implementation with the necessary features.
+# If you decide to use my implementation you can ignore this part.
+#
+#
+# - The computer needs to implement memory much /larger/ than the set of initial commands.
+#   Any memory address not part of the initial commands can be assumed to be initialized to 0.
+#   (only positive addresses are valid).
+# - You need to support a new parameter mode, 'relative mode', denoted as mode 2 in the 'mode' part
+#   of the instructions.
+#   Relative mode is similar to position mode (the first access mode you implemented). However, 
+#   parameters in relative mode count not from 0, but from a value called 'relative offset'. 
+#   When the computer is initialized, the relative offset is initialized to 0, and as long as it remains
+#   0 relative mode and position mode are identical.
+#   In general though parameters in relative mode address the memory location at 'relative offset + parameter value'.
+#   EXAMPLE: if the relative offset is 50, the mode is 2, and the value you read from memory is 7 you should 
+#     retrieve data from the memory address 57.
+#     Equally, if you read -7, you should retrieve data from the memory address 43.
+#   This applies to both read- and write operations.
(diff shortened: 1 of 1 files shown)


atrejut:
- as far as I can tell you call `move_paddle` every time you got three value from the computer -- even if that particular set of values was not related to ball- or paddle-position at all...
- within `move_paddle` you /think/ you got a `ball_coordinate` as argument, but really what you pass is just the last value you got from the int-computer? which actually is a tile type I think, and not a position of anything at all?
- and this leads to you making the wrong move: the first time you are asked to provide an input the real x-coordinates are pad=21, ball=19, so you should be going left -- but you decide to go right, probably just because the last input to `move_paddle` happened to make you go right accidentally.

**general advice**:
you are trying to do something with the computers output every time you receive a new value. I would recommend just storing the value you get at that point, because you don't actually need to parse it until you are asked to provide a direction. at that point, go through all the stored outputs, figure out where ball and paddle are, and go the right direction...

Exercises/JanMerz/exercises_07.py
```diff
@@ -0,0 +1,255 @@
...
+        return paddle_direction
+
+    def paint_picture(self):
+
+        self.ax.imshow(self.matrix)
+        plt.draw()
+        plt.pause(0.000001)
+        self.ax.clear()
+
+root_dir = pathlib.Path().absolute().parents[1]
```
DrKeks1010, line 246: Change this to:
root_dir = pathlib.Path(__file__).absolute().parents[2]

This way, the code will use this file as an anchor when searching for breakout_commands. This way, everybody can execute this file and it doesnt matter, where they execute it from

Exercises/JanMerz/exercises_07.py
```diff
@@ -0,0 +1,255 @@
...
+with open(root_dir / "data" / "breakout_commands.txt", "r") as commands:
+    lines = commands.read().splitlines()
+    for element in range(len(lines)):
+        lines[element] = int(lines[element])
+
+run_obj = RunGame()
+computer = IntComputer(run_obj.input_getter, run_obj.output_method)
+picture_pixels = []
```
DrKeks1010, line 254: unnecessary assignment, please remove

Exercises/JanMerz/exercises_07.py
```diff
@@ -0,0 +1,255 @@
...
+    zaehler = 0
+    relevant_tiles = 0
+    x_coordinate = 0
+    y_coordinate = 0
+    ball_coordinate = 0
+    matrix = np.zeros([23, 43])
+    figure, ax = plt.subplots()
+
+    def output_method(self, computer_return):
+
```
DrKeks1010, line 178: Not Mandatory: (improved readability and code structure)

you might want to consider splitting this method into two:
one that fetches all the inputs and stores them (might add into deque with size limit of 3)
And another one, that will read all 3 bits of data, once all have been collected. This way this method will be much easier to read. This will have the added benefit of you not having to store the x_coordinate, y_coordinate, zaehler, relevant_tiles and ignore_value. As you will always have all 3 required values when entering this method.

Exercises/JanMerz/exercises_07.py
```diff
@@ -0,0 +1,255 @@
...
+            opcode_function = self.function_map[opcode]
+            opcode_function(modes)
+
+        print('shutting down...')
+
+class RunGame:
+
+    ignore_value = False
+    painted_main_screen = False
+    zaehler = 0
```
DrKeks1010, line 169: Not Mandatory: (code consistency)

Change zaehler to counter / change everything into german

Exercises/JanMerz/exercises_07.py
```diff
@@ -0,0 +1,255 @@
...
+    relevant_tiles = 0
+    x_coordinate = 0
+    y_coordinate = 0
+    ball_coordinate = 0
+    matrix = np.zeros([23, 43])
+    figure, ax = plt.subplots()
+
+    def output_method(self, computer_return):
+
+        if (computer_return == -1 or self.ignore_value):
```
DrKeks1010, line 179: Not Mandatory: (code style)

brackets aren't required and thus should be removed

Exercises/JanMerz/exercises_07.py
```diff
@@ -0,0 +1,255 @@
...
+
+            if(self.zaehler == 0 and not self.painted_main_screen):
```
DrKeks1010, line 181: Same as above

Exercises/JanMerz/exercises_07.py
```diff
@@ -0,0 +1,255 @@
...
+                self.paint_picture()
+                self.painted_main_screen = True
+
+            self.zaehler += 1
+
+            if(self.zaehler < 3):
```
DrKeks1010, line 187: Same as above

Exercises/JanMerz/exercises_07.py
```diff
@@ -0,0 +1,255 @@
...
+                self.ignore_value = True
+                return
+            elif(self.zaehler == 3):
```
DrKeks1010, line 190: Same as above

Exercises/JanMerz/exercises_07.py
```diff
@@ -0,0 +1,255 @@
...
+        match self.relevant_tiles:
+            case 1: 
+                self.x_coordinate = computer_return
+            case 2: 
+                self.y_coordinate = computer_return
+            case 3: 
+                self.change_matrix(computer_return)
+                self.relevant_tiles = 0
+
+                if(computer_return == 4):
```
DrKeks1010, line 210: Same as above

Exercises/JanMerz/exercises_07.py
```diff
@@ -0,0 +1,255 @@
...
+                    self.ball_coordinate = self.x_coordinate
+                    if(self.painted_main_screen):
```
DrKeks1010, line 212: Same here

Exercises/JanMerz/exercises_07.py
```diff
@@ -0,0 +1,255 @@
...
+                        self.paint_picture()
+            
+    
+    def move_paddle(self):
+        paddle_coordinate = 0
+
+        for index in range(43):
+            if(self.matrix[21, index] == 3):
```
DrKeks1010, line 220: And here

Exercises/JanMerz/exercises_07.py
```diff
@@ -0,0 +1,255 @@
...
+                paddle_coordinate = index
+                break
+
+        if(paddle_coordinate < self.ball_coordinate):
```
DrKeks1010, line 224: And here

Exercises/JanMerz/exercises_07.py
```diff
@@ -0,0 +1,255 @@
...
+                paddle_direction = 1
+        elif(paddle_coordinate > self.ball_coordinate):
```
DrKeks1010, line 226: And here a last time

DrKeks1010 (changes requested):
Overall a great solution.
Still minor tweaks are required (first two comments), shouldn't take more than 5 minutes, as your solution works overall and is a great solution. 

Other changes aren't required, but will help with code style, readability and consistency.

Exercises/JanMerz/exercises_07.py
TrueEsel: while I do see your point and believe that you're right, I do not have the time to make this change this week

DrKeks1010 (approved):
All problems fixed. Great Solution

PR #164, opened 2023-11-20
Paul seidel exercise06 (1778099,X)

paulseidel08:
Finally finished part1. Unfortunately, I spent so much time on debugging, I coudn't make part2. Hours spent: about 7h

But part1 should behave like it's supposted to.

Now the code is importing the input file via pathlib


Exercises/PaulSeidel/PSeidel_exercises_06.py
```diff
@@ -0,0 +1,224 @@
...
+#terminal_list = ["$ cd /", "$ ls", "dir subdir1", "123000 subfile1.bs", "dir subdir2", "$ cd subdir1", "$ ls", "234 subsubfile1.fg", "dir subsubdir1"]
+#result: 234 (correct. Because the dir "/" is bigger than 100000, only subdir1 counts, which's size is 234)
+
+#terminal_list = ["$ cd /", "$ ls", "dir subdir1", "123 subfile1.bs", "dir subdir2", "$ cd subdir1", "$ ls", "234 subsubfile1.fg", "dir subsubdir1", "$ cd ..", "$ cd subdir2", "$ ls", "dir subsubdir1", "300 subsubfile1.fg"]
+#result: 1191 (correct)
+
+#terminal_list = ["$ cd /", "$ ls", "dir subdir1", "123 subfile1.bs", "dir subdir2", "$ cd subdir1", "$ ls", "234 subsubfile1.fg", "dir subsubdir1", "$ cd ..", "$ cd subdir2", "$ ls", "dir subsubdir1", "300 subsubfile1.fg", "$ cd subsubdir1", "$ ls", "100001 file3", "$ cd ..", "$ cd ..", "$ cd subdir1", "$ cd subsubdir1", "$ ls", "200 file1", "300 file2", "$ cd ..", "$ cd .."]
+#_____________________________________________________________________________________________________________________________________
+
+with open("./data/terminal_record.txt", "r") as input_file:
```
fabischw, line 96: While I like the fact that you're trying to isolate the file into a data directory, the code in it's current state won't run unless I manipulate the source code or create a directory called data in your working directory and copying the record over from the main/data directory. Consider using relative paths or using a library like pathlib to just read the file from (relative path): ```./../../data/terminal_record.txt ``` .

Exercises/PaulSeidel/PSeidel_exercises_06.py
```diff
@@ -0,0 +1,224 @@
...
+            #former_change_of_dir = line[2]
+                
+        elif line[1] == "ls":   #can be ignored -> lines starting with dir or a numer (for a file) can only exist behind a "$ ls".
+            pass
+
+        elif line[0] == "dir":
+            #full_path = list(current_path.append(f"{parents_path}"))   #the full path contains the path of the dir including it's name
+            #if current_path not in folder_structure:    #to make sure, no existing dir is beeing overwritten.
+            name_of_new_dir = (f"{line[1]}",)
+            folder_structure[current_path + name_of_new_dir] = []    #adding the dir to the dict with an empty list as value and the full path-list as string as key.
```
thecodingcube77, line 148: `collections.defaultdict` can save you this initialization with an empty list.

Exercises/PaulSeidel/PSeidel_exercises_06.py
```diff
@@ -0,0 +1,224 @@
...
+
+            temp = folder_structure[current_path]
+            temp.append(("dir", f"{line[1]}"))
+            folder_structure[current_path] = temp  #adding the dir as a tuple ("dir", "<name>") to its parent-dir
+
+        else:   #element has to be a file
+            temp = folder_structure[current_path]
+            temp.append((int(line[0]),f"{line[1]}"))   #adding a tuple with size and name of the file to the current dir
+            folder_structure[current_path] = temp
```
thecodingcube77, line 157: saving individual file names and sizes is not really needed in this task as you only need the sizes of whole directories but could maybe be useful sometime.

Exercises/PaulSeidel/PSeidel_exercises_06.py
```diff
@@ -0,0 +1,224 @@
...
+
+def recursive_calculate_size(current_directory = ("/",)):
```
thecodingcube77, line 159: This whole function could be done inside the function above by adding the adding the size of a currently encountered file to the current as well as all upstream directories. This would save a bit of work in my opinion.
Could be done like this:
```py
if current_line_has_file:                             # if the element currently listed is a file
    for path in cwd_and_upstream_paths(current_path): # loop through all parent paths from the current directory to the root directory
        dir_sizes_dict[path] += current_file_size     # add the size of the currently listed file to the directory
```
This would be done in the last `else` in the `create_folder_structure()` function.
I would create a generator for `cwd_and_upstream_paths()`  which takes the `current_dir` as an argument. Using `pathlib` for the `current_dir` makes the creation of the generator really easy.

Exercises/PaulSeidel/PSeidel_exercises_06.py
```diff
@@ -0,0 +1,224 @@
...
+            temp = dir_sizes[current_directory]
+            temp += subelement[0]
+            dir_sizes[current_directory] = temp
+
+        elif subelement[0] == "dir":    #execute the function with the subdirectory (recursion)
+            new_path_part = (f"{subelement[1]}",)
+            new_path = current_directory + new_path_part
+            subelement_size = recursive_calculate_size(new_path)
+            temp = dir_sizes[current_directory] + subelement_size
+            dir_sizes[current_directory] = temp
```
thecodingcube77, line 177: This being so complicated is due to you creating a very complicated system to store your data.

Exercises/PaulSeidel/PSeidel_exercises_06.py
```diff
@@ -0,0 +1,224 @@
...
+    recursive_calculate_size()
+    return dir_sizes
+
+def sum_dirs_of_criteria(dictionary: dict):
+    """Sums sizes of all directories in the given dictionary which sizes are at most 100000 and returns the total sum"""
+    sum = 0
+    for dir_name in dictionary:
+        if dictionary[dir_name] <= 100000:
+            sum += dictionary[dir_name]
+    return sum
```
thecodingcube77, line 194: You could just write
```py
return sum([dir_size for dir_size in dictionary.values() if dir_size <= 100000])
```

Exercises/PaulSeidel/PSeidel_exercises_06.py
```diff
@@ -0,0 +1,224 @@
...
+    return dir_sizes[current_directory]
+
+def run_and_return_dict():
+    """runs the upper functions to analyze the input file into the folder structure and calculate the size of the dir "/" and all of it's subdirs
+    and returns the directory of sizes. That's required by the exercise."""
+    create_folder_structure()
+    recursive_calculate_size()
+    return dir_sizes
+
+def sum_dirs_of_criteria(dictionary: dict):
```
thecodingcube77, line 188: I would not name it dictionary. What information does the dictionary hold?  I don't know from reading this.

Exercises/PaulSeidel/PSeidel_exercises_06.py
```diff
@@ -0,0 +1,224 @@
...
+    """Sums sizes of all directories in the given dictionary which sizes are at most 100000 and returns the total sum"""
+    sum = 0
+    for dir_name in dictionary:
```
thecodingcube77, line 191: You can just loop over the values of a dict by using `for value in dictionary.values()`

Exercises/PaulSeidel/PSeidel_exercises_06.py
```diff
@@ -0,0 +1,224 @@
...
+# Delete directory `d`, which would increase unused space by 24933642.
+# Delete directory `/`, which would increase unused space by 48381165.
+# Directories `e` and `a` are both too small; deleting them would not free up enough space.
+# However, directories `d` and `/` are both big enough! 
+# Between these, choose the smallest: `d`, increasing unused space by 24933642.
+# 
+# Mention the total size of the deleted directory in the PR.
+# 
+# You can find the /actual/ input to both parts in data/terminal_record.txt
+###############################################################################
```
thecodingcube77, line 224: Quick suggestion for part 2:
```py
def find_dir_to_delete(needed_size: int, file_system_size: int, file_sizes_dict: dict):
    # calculate the space you need to free
    size_of_dir_to_delete = min(dir_size for dir_size in dir_sizes_dict.values() if dir_size >= needed_space) # find the smallest dir size which would free up enough space
    index_of_dir = list(dir_sizes_dict.values()).index(size_of_dir_to_delete) # find the index of the determined directory by the determined size
    path_of_dir_to_delete = list(dir_sizes_dict.keys())[index_of_dir]         # get the key(=path) from the list of keys as the index from the values directly translates to the keys
    return path_of_dir_to_delete
```
I have to note that this way of interacting with a dict is really inefficient as a dict is not made to be used the other way around.

Exercises/PaulSeidel/PSeidel_exercises_06.py
```diff
@@ -0,0 +1,224 @@
...
+
+def sum_dirs_of_criteria(dictionary: dict):
+    """Sums sizes of all directories in the given dictionary which sizes are at most 100000 and returns the total sum"""
+    sum = 0
+    for dir_name in dictionary:
+        if dictionary[dir_name] <= 100000:
+            sum += dictionary[dir_name]
+    return sum
+
+print(sum_dirs_of_criteria(run_and_return_dict()))
```
thecodingcube77, line 196: You should save the `dir_dict` in a variable to reuse it in task 2 (if you decide to do it).

thecodingcube77 (changes requested):
You don't make it easy for yourself.

thecodingcube77:
Also, when creating new branches:
first: `git checkout main`
second: `git pull`
third: `git checkout -b new_branch_name`

fabischw (approved):
Path problems were fixed. Approving this as you stated you do not want to spend any more time on the assignement.

thecodingcube77 (approved):
I'll approve as well due to the time you already spent on the exercise.

fabischw:
If you want to merge the code in it's current state, I recommend adding a docstring at the top of the python document explaining why it's not finished and what exactly is not finished to avoid future confusion

PR #165, opened 2023-11-22
Julius Dralle Task 7. (17159)

RubelOS:
It is working.
PicturePart1

https://github.com/dhbw-ma-ppp/ppp-2023/assets/35865838/bd58ea20-8960-4e84-922c-817258ac6426


Exercises/JuliusDralle/exercises_07.py
```diff
@@ -0,0 +1,239 @@
...
+        plt.clf()
+        
+
+
+
+with open(path_dir_data / "breakout_commands.txt", "r") as input_file:
+    input_lines = input_file.read().splitlines()
+    input_lines_int = [int(x) for x in input_lines]
+    input_lines = input_lines_int
+    input_lines_int = None
```
thecodingcube77, line 228:
```py
del input_lines_int
```
Will remove the variable and free it's space.
To just create one list you could just
```py
input_lines = [int(x) for x in input_file.read().splitlines()]
```
Should do exactly the same.

fabischw (changes requested):
Please remove the data directory or revert the file to it's original state. In the current PR, you change the breakout_commands.txt file. You can either just remove this in the PR or write wrapper function for both exercises with the second one replacing the value at index 0 so the original file is preserved in it's original state.

fabischw (approved):
Problem was adressed. You might want to consider adding a requirements.txt so users and reviewers can easily view and install dependencies. This is optional for now in my opinion as every student of the class should now have numpy and matplotlib installed.

fabischw:

PR #167, opened 2023-11-22
Philipp Kirschke exercises_07.py UNFINISHED

The-Linux-User:
This is like the 27th version, first I tried it without help, then with Google and then with the help of ChatGPT and I cannot get this working. I spent around 3 1/2 hours trying and I am positive that more time will not help, so I unfortunately have to commit my code in a "not working" state. This version is the one where I got the furthest but all of the tiles are in the wrong place and I just cannot get that fixed (neither can ChatGPT), every fix I tried either does nothing at all or breaks the rest of the code


Exercises/Philipp Kirschke/exercises_07.py
```diff
@@ -0,0 +1,167 @@
...
+#   the offset increases by the value of the parameter (or decreases if that value is negative).
+import os
+from pathlib import Path
+import tkinter as tk
+from int_computer import IntComputer
+
+import os
+from pathlib import Path
+import tkinter as tk
+from int_computer import IntComputer
```
GeFritz, line 79: It seems like you did not add the file `int_computer` to this pull request. Please add it so debugging is possible. Did you use Atrejus int computer?. You still have to add it to the directory `ppp-2023\Exercises\Philipp_Kirschke` if you import it like that.

Exercises/Philipp Kirschke/exercises_07.py
```diff
@@ -0,0 +1,167 @@
...
+            self.current_y += 1
+
+    def get_input(self):
+        # Logic to control the paddle based on the ball's position
+        # You need to implement this based on the task description
+        return 0  # For now, the paddle remains in position
+
+def main():
+    # Read the file
+    working_dir = Path(os.getcwd()).parent.absolute().parent.absolute()
```
GeFritz, line 154: This path is working directory dependent. `pathlib` is preferred. The problem is that i could run your python file on `ppp-2023` directory and this code would not work.

Exercises/Philipp Kirschke/exercises_07.py
```diff
@@ -0,0 +1,167 @@
...
+
+        # Create a variable to store the output
+        self.current_output = None
+
+        # Initialize x and y coordinates
+        self.current_x = 0
+        self.current_y = 0
+
+        # Set the canvas width (adjust this based on your canvas size)
+        self.canvas_width = 10
```
GeFritz, line 98: The width has to be 43.

Exercises/Philipp Kirschke/exercises_07.py
```diff
@@ -0,0 +1,167 @@
...
+        def output_wrapper(output):
+            self.current_output = output
+            self.update_display(output)
+
+        # Create an instance of IntComputer with the wrapped output_collector
+        self.computer = IntComputer(self.get_input, output_wrapper)
+        self.computer.run(program)
+
+    def update_display(self, tile_type):
+        # Map tile types to colors (you can customize this mapping)
```
GeFritz, line 110: Add logic e.g. using a triplet counter that only adds the first, fourth, .. time. Using modulo for instance.
`if not (self.triplet_counter+1) % 3:`
Indent the rest and include  `self.triplet_counter += 1` at the end of the method.

GeFritz (changes requested):
The error in your code is that a rectangle is drawn for every call of the output getter function. The output should only be drawn every third call as the computer prints out the x coordinate, the y coordinate and then thy type. Manually going though the rows is therefore not strictly necessary. If mentioned changed are implemented your code will look like this:
image

Exercises/Philipp Kirschke/exercises_07.py
```diff
@@ -0,0 +1,167 @@
...
+import tkinter as tk
+from int_computer import IntComputer
+
+class ArcadeGame:
+    def __init__(self, master, program):
+        self.master = master
+        master.title("Advent of Code 2019 - Day 13 Arcade Game")
+
+        # Create a canvas to draw the game
+        self.canvas = tk.Canvas(master, width=400, height=400)
```
GeFritz, line 87: This has to be changed to `self.canvas = tk.Canvas(master, width=850, height=400)`.

The-Linux-User:
Bildschirmfoto 2023-11-24 um 08 44 28

Task 1 working now, thanks @GeFritz!

Exercises/Philipp Kirschke/exercises_07.py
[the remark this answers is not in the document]
The-Linux-User: Removed the redundant code

The-Linux-User: Will change that asap

Exercises/Philipp Kirschke/exercises_07.py
The-Linux-User: Changed to 43

Exercises/Philipp Kirschke/exercises_07.py
The-Linux-User: Changed the width to 800

The-Linux-User:
> The error in your code is that a rectangle is drawn for every call of the output getter function. The output should only be drawn every third call as the computer prints out the x coordinate, the y coordinate, and then the type. Manually going through the rows is therefore not strictly necessary. If the below mentioned changes are implemented, your result will look like this: image

Thanks a lot, now it works.

Exercises/Philipp Kirschke/exercises_07.py
The-Linux-User: Using pathlib now

Exercises/Philipp Kirschke/exercises_07.py
The-Linux-User: Will do that next time, can't seem to figure out how to upload additional files to the pull request, if I just do a commit to the current branch it doesn't show up on GitHub

PR #171, opened 2023-11-23
solution for exercise 7 (17159)

christianrietzsch:
i will add comments later

### DIFF
--- a/Exercises/ChristianRietzsch/exercises_07.py
+++ b/Exercises/ChristianRietzsch/exercises_07.py
@@ -1,0 +1,155 @@
+import operator
+import os
+
+global three_values, x, y, type, high_score, c_input, buffer
+three_values, x, y, type, buffer  = [], [], [], [], []
+high_score, c_input = 0, 0
+
+class IntComputer:
+    def __init__(self, input_getter, output_collector):
+        def get_input(modes):
+            target_mode = modes % 10
+            if target_mode == 2:
+                target = self.memory.get(self.ip, 0) + self.relative_mode_offset
+            else:
+                target = self.memory.get(self.ip, 0)
+            self.memory[target] = input_getter()
+            self.ip += 1
+
+        def write_output(modes):
+            x = self.get_function_arguments(modes, 1)[0]
+            output_collector(x)
+
+        def set_offset(modes):
+            x = self.get_function_arguments(modes, 1)[0]
+            self.relative_mode_offset += x
+
+        self.function_map = {
+            1: self._make_register_setter(operator.add),
+            2: self._make_register_setter(operator.mul),
+            3: get_input,
+            4: write_output,
+            5: self._make_ip_setter(operator.ne, 0),
+            6: self._make_ip_setter(operator.eq, 0),
+            7: self._make_register_setter(operator.lt),
+            8: self._make_register_setter(operator.eq),
+            9: set_offset,
+        }
+
+    def _make_register_setter(self, func):
+        # make generic function that sets a register
+        def f(modes):
+            x, y = self.get_function_arguments(modes, 2)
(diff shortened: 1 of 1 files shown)


Exercises/ChristianRietzsch/exercises_07.py
```diff
@@ -0,0 +1,163 @@
+import operator
+import os
+import sys
+
+global three_values, x, y, type, high_score, c_input, buffer
```
FritzRa, line 5: usage of global variables not recommended instead use returns or create a class structure for running the game

Exercises/ChristianRietzsch/exercises_07.py
```diff
@@ -0,0 +1,163 @@
...
+    def run(self, data):
+        self.memory = {i: v for i, v in enumerate(data)}
+        self.ip = 0
+        self.relative_mode_offset = 0
+
+        while True:
+            opcode, modes = self.split_command_and_modes()
+            if opcode == 99:
+                global high_score
+                print(f"\n[Highscore]: {high_score}")
```
FritzRa, line 91: while running the code on manual mode the highscore was not updated (stayed at zero), however when running the game on automatic mode the end highscore was correct (17159)

Exercises/ChristianRietzsch/exercises_07.py
```diff
@@ -0,0 +1,163 @@
...
+                break
+            # calculate function arguments
+            opcode_function = self.function_map[opcode]
+            opcode_function(modes)
+
+
+def input_getter():
+    graphical_output()
+    if("computer" in game_type):
```
FritzRa, line 100: nice implementation of switching between gamemodes (inputs and automatic)

Exercises/ChristianRietzsch/exercises_07.py
```diff
@@ -0,0 +1,163 @@
...
+            case 3:
+                type.append("_")
+            case 4:
+                type.append("x")
+            case _:
+                type.append(" ")
+
+        three_values = []
+
+def graphical_output():
```
FritzRa, line 130: interesting concept to print the graphical ouput in terminal

Exercises/ChristianRietzsch/exercises_07.py
```diff
@@ -0,0 +1,163 @@
...
+    if(x_point == x_platform):
+        c_input = 0
+    elif(x_point > x_platform):
+        c_input = 1
+    else:
+        c_input = -1
+
+ic = IntComputer(input_getter, output_collector)
+
+with open(".."+os.sep+".."+os.sep+"data"+os.sep+"breakout_commands.txt", "r") as file:
```
FritzRa, line 159: os.path did not work on vscode for me (had to hardcode filepath), implementation of Pathlib recommended

Exercises/ChristianRietzsch/exercises_07.py
```diff
@@ -0,0 +1,163 @@
...
+    lines = [line.split("\n")[0] for line in file.readlines()]
+    [lines.append("0") for num in range(1000)]
```
FritzRa, line 161: predetermined number of zeros added, not the best solution, could also be dynamic. For instance if a new element is added to the list, the program checks if its index is outside of the list and if True it fills the list with zeros until it reaches that index

FritzRa (changes requested):
All in all mostly nice solution to the task. There where some issue when trying to run the program in Vscode, for instance that the file was not found using the os.sep. Another issue is that the highscore was not updated after destroying a block [always stays at zero](in manual mode). Lastly one weird bug was that the program only ran in debug mode in VScode, else a syntax error at match case got raised. I could not find out why. 

Requested changes:
-working highscore in manual mode

PR #172, opened 2023-11-23
Laura Lösung Woche 7 (17159)

s231365:
VL7 Bild

### DIFF
--- a/Exercises/LauravanUffelt/VL7 Lösungen.py
+++ b/Exercises/LauravanUffelt/VL7 Lösungen.py
@@ -1,0 +1,142 @@
+# to solve todays exercise you will need a fully functional int-computer
+# a fully functional int-computer has some additional features compared to the
+# last one you implemented.
+# you can either use my implementation of an int-computer (see AtrejuTauschinsky/int_computer.py)
+# or you can extend your own int-computer with the necessary features. If you want to extend your own
+# the features you need to implement will be described at the bottom.
+#
+#
+# We will run 'breakout' -- the arcade game -- on our simulated computer.
+# (https://en.wikipedia.org/wiki/Breakout_(video_game))
+# The code for the computer will be provided under data/breakout_commands.txt
+# the code will produce outputs in triplets. every triplet that is output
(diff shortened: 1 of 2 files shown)


UtkuDHBW (approved):
Well structured code overall makes it easier to understand, looks good to me!

Exercises/LauravanUffelt/VL7 Lösungen.py
```diff
@@ -0,0 +1,142 @@
...
+        return liste
+
+
+def analyse():
+    global paddle_pos
+    global ball_pos
+    global score
+    global frame_data
+    for index in range(0, len(frame_data), 3):  # goes through the commands (Schrittweite 3)
+        x, y, type = frame_data[index], frame_data[index+1], frame_data[index+2] # saves the commands in variables
```
UtkuDHBW, line 102: You can shorten the line 102, not needed but its nice:

```py
x, y, type = frame_data[index:index+3]
```

UtkuDHBW: Also you shouldn`t use variables that are also a builtins like ```type```, instead you should use something else that's still clear to understand like for example "command_type" or something else. A underscore like "type_" would also be enough.

Exercises/LauravanUffelt/VL7 Lösungen.py
```diff
@@ -0,0 +1,142 @@
...
+            liste.append(line)  # appends evey command to a list
+        liste[0] = 2  # changes the first command for task 2
+        return liste
+
+
+def analyse():
+    global paddle_pos
+    global ball_pos
+    global score
+    global frame_data
```
UtkuDHBW, line 100: Here i recommend that it should be a common practice to not use any global variables since this makes debugging and also understanding the code harder. I recommend to generally use function parameters or a class.

As far as I know there are very few scenarios where a global variable would be better

CelinaRiegel (approved):
Looks good. You could rename the variable type to something else, to not confuse it with the build-in.

fabischw:

PR #173, opened 2023-11-23
Fabian woche7

fabischw:
Solution for week 7, commit ea25c2bc30db040f9c3fcc48f69c005dbed5c494 was done during pair programming with @LeMixer.


Exercises/AtrejuTauschinsky/example.py
```diff
@@ -1,2 +0,0 @@
-if __name__ == '__main__':
-    print('Hello World!')
```
GeFritz, line 2: Why are you removing this?

Exercises/FabianSchwarz/exercises_07.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+
+
+
+
+
+class no_bloat_output:
+    """
+    fancy terminal output for the game that has 0% bloat
```
RubelOS, line 215: Okay, but where is it?

RubelOS (approved):
Looks really polished.

Exercises/AtrejuTauschinsky/example.py
RubelOS: It was a main feature!!

Exercises/FabianSchwarz/exercises_07.py
fabischw: This class was deprecated as I decided to only provide one interface instead of two different ones so no seperate class is needed.

Exercises/AtrejuTauschinsky/example.py
fabischw: > Why are you removing this?

Was an accident, has been reverted.

fabischw:
grafik

PR #177, opened 2023-11-23
Celina 7

CelinaRiegel:
Part 1 is now working.
Figure_1


TomKnoden:
Try to reduce the use of global variables in your code. Instead, consider passing variables as parameters to functions.
For example in line 56 and 57.

Exercises/Celina/exercises_07.py
```diff
@@ -0,0 +1,141 @@
...
+import numpy as np
+
+commands=[]
+output = []
+
+paddle_pos=None
+ball_pos=None
+tiles= np.zeros((23,43))
+
+with open("breakout_commands.txt", "r") as data:
```
christianrietzsch, line 43: you should use `../../data/breakout_commands.txt` as the path
if you want to directly jump to the game with inputs
(which you should, since you have the graphical output function in the input collector)
add `commands[0] = 2` after `commands.append()`

Exercises/Celina/exercises_07.py
```diff
@@ -0,0 +1,141 @@
...
+    0: "white",
+    1: "green",
+    2: "blue",
+    3: "red",
+    4: "yellow"
+}
+
+def make_type(x,y,types):
+    global paddle_pos
+    global ball_pos
```
christianrietzsch, line 57: no global variables needed

Exercises/Celina/exercises_07.py
```diff
@@ -0,0 +1,141 @@
...
+    tiles[x,y]=types
```
christianrietzsch, line 58: you need to switch `x` and `y` to match the axis with the game: `tiles[y,x]`
you also have to check for types that have other values than 1,2,3,4 because you do not filter
the output for highscore-values: `= types if types in [1,2,3,4] else 0`

Exercises/Celina/exercises_07.py
```diff
@@ -0,0 +1,141 @@
...
+    elif types==4:
+        #ax.bar(x,y,1,y,color="yellow")
+        ball_pos=x
+        
+        
+def breakout():
+    #fig, ax = plt.subplots()
+    for element in output:
+        #print(output[element],output[element+1],output[element+2])
+        x = output[element]
```
christianrietzsch, line 71: you are using the element as an index

Exercises/Celina/exercises_07.py
```diff
@@ -0,0 +1,141 @@
...
+        y = output[element+1]
+        make_type(x,y,output[element+2])
+        element+=3
```
christianrietzsch, line 74: you are incrementing the respective element of output and not the index
(not needed if you change your code to the for loop in my comment)

Exercises/Celina/exercises_07.py
```diff
@@ -0,0 +1,141 @@
...
+        #ax.bar(x,y,1,y,color="red")
+        paddle_pos=x
+    elif types==4:
+        #ax.bar(x,y,1,y,color="yellow")
+        ball_pos=x
+        
+        
+def breakout():
+    #fig, ax = plt.subplots()
+    for element in output:
```
christianrietzsch, line 69: use `for index in range(len(output))[::3]:`
instead and replace element with index

Exercises/Celina/exercises_07.py
```diff
@@ -0,0 +1,141 @@
...
+    if ball_pos > paddle_pos:
+        return 1
+    elif ball_pos < paddle_pos:
+        return -1
+    else:
+        return 0
+    
+MyIntComputer = int_computer.IntComputer(input_getter,output_collector)
+MyIntComputer.run(commands)
+breakout()
```
christianrietzsch, line 96: not needed if you change directly to the game (task 2)

Exercises/Celina/exercises_07.py
```diff
@@ -0,0 +1,141 @@
...
+        y = output[element+1]
+        make_type(x,y,output[element+2])
+        element+=3
+    #plt.show()
+    
+def output_collector(value):
+    output.append(value)
+    
+def input_getter():
+    global output
```
christianrietzsch, line 81: global variable not needed

Exercises/Celina/exercises_07.py
```diff
@@ -0,0 +1,141 @@
...
+    breakout()
+    output=[]
+    fig, ax = plt.subplots()
```
christianrietzsch, line 84: not needed: just use `plt.imshow()`
before you refresh the screen you could clear it with `plt.cla()`

Exercises/Celina/exercises_07.py
```diff
@@ -0,0 +1,141 @@
...
+    ax.imshow(tiles)
+    plt.show()
```
christianrietzsch, line 86: this makes the screen static
use `plt.show(block=False)` and `plt.pause(0.0001)` instead

christianrietzsch (changes requested):
the game is now working
i would recommend using `return int(input())` in line 86 to play the game manually
to make a screenshot for task 1, because the last minute of the computer playing 
uses a lot of resources
(i did not play it to the end with graphical output enabled, but tested with graphical output disabled that the game works!)

christianrietzsch (approved):
LGTM

fabischw:

PR #180, opened 2023-11-23
Reupload zu Woche 02

TomKnoden:
As stated in the title, this PR simply is a reupload of my solution for week 01. There have been problems concerning the former branch in which the file was located in. These problems have been fixed.


RubelOS (approved):
Good to go

PR #185, opened 2023-11-25
Aufgaben 8 Jan und Jasmin (1. Version nur für Abgabe)

not-too-deranged:
any future changes will only be committed on "Loesung_08_Jasmin_Steidl_Jan_Merz"


paulseidel08 (approved):
Nice solution!
Very illustrative graphs with different colors, I like it!

However, the complexity of code for your graphs could have been way less for this task and your could have written more for analysis.

PR #186, opened 2023-11-28
Paul seidel exercise08

paulseidel08:
Did it with Jupyter Note


GeFritz (changes requested):
You seem to also commit files from previous weeks. Please adjust.

GeFritz (approved):
Nice work!

fabischw (commented):
Nice work, but consider naming it something like 'age_group' instead of 'new_age' as this can be quite confusing.

PR #187, opened 2023-11-28
Gerrit Fritz Aufgabe 8

GeFritz:
Done with @FritzRa


The-Linux-User (approved):
Really nice selection of data!

PR #189, opened 2023-11-28
Aufgaben Woche 8 Raphael Fritz

FritzRa:
pair programming with @GeFritz


s231365 (approved):
I like your representation of the numbers. Especially the first Graph for the fare.

PR #190, opened 2023-11-28
Aufgaben zu Woche 07

TomKnoden:
In order to run the code it is required for you to have Matplotlib installed on your computer.
You could install Matplotlib by typing "pip install matplotlib" into your terminal, for example.


Exercises/Tom_Knoden/Woche07_Aufgaben.py
```diff
@@ -0,0 +1,59 @@
...
+        for coord, tile_type in self.screen.items():
+            x, y = coord
+            grid[y - min_y, x - min_x] = tile_type
+
+        plt.imshow(grid, cmap='viridis', interpolation='none', aspect='equal')
+        plt.title(f'Score: {self.score}')
+        plt.show()
+
+# Load the program from the file using PathLib
+input_file_path = Path('data/breakout_commands.txt')
```
fabischw, line 45: Thanks for using pathlib. However the path is still dependant on the current working directory of the reviewer. You should use something like this:
```py
here = pathlib.Path(__file__).parent
exercise_dir = here.parent

with open(data_dir / "breakout_commands.txt","r") as file:
    data = file.readlines()

```

instead.
I also want to point out that you deleted Re-upload files in your PR; is this intended ?

fabischw (changes requested):
Check comments below. Reason for the CR is the file import path.

PR #192, opened 2023-11-28
Aufgaben zu Woche08

TomKnoden:
Just like in my solution for week 07, you'll need to have Matplotlib installed on your device in order to run this code.
Additionally, you'll need to have pandas installed on your device.
If you haven't already, you can use the following commands to install these libraries in your terminal:

"pip install pandas"
pip install matplotlib"


Exercises/Tom_Knoden/Woche08_Aufgaben_Version2.py
```diff
@@ -0,0 +1,72 @@
...
+# calculations/plots/... might be interesting to analyse. You will find very many such analyses online
+# for the titanic data, but I would recommend trying to think of your own ideas first, and only later
+# look for more inspiration online.
+
+import pandas as pd
+import matplotlib.pyplot as plt
+from pathlib import Path
+
+input_file_path = Path('data/titanic.csv')
+titanic_data = pd.read_csv(input_file_path)
```
TrueEsel, line 27: this way of reading in the file is not working for me. you could use something like
```suggestion
root_dir = pathlib.Path(__file__).absolute().parents[2]
with open(root_dir / "data" / "titanic.csv", "r") as ifile:
    titanic_data = pd.read_csv(ifile)
```
instead.

TrueEsel (changes requested):
the task clearly states 
> The output should not just be prints of numbers or figures, but also some explanatory text of
what you have been analysing, and which conclusions you can draw from the different steps of your analysis.
Therefore you did not complete it. Furthermore you are only showing absolute numbers and not the percentage of surviving passengers per group, which I believe would make more sense.

TrueEsel (changes requested):
the path is fine now. However, you still haven't drawn conclusions from the different steps of your analysis which is a part of the task

PR #193, opened 2023-11-29
Henriette_Neumann.exercise.7

HenrietteNeumann:
Late because sick

### DIFF
--- a/exercise_7.py
+++ b/exercise_7.py
@@ -1,0 +1,71 @@
+from matplotlib import pyplot as plt
+import pathlib
+import numpy as np
+import Int_Computeromputer
+
+output = []
+matrix = np.zeros([23, 43])
+ball_pos = 0
+paddle_pos = 0
+score = 0
+first_frame = True
+
+
+def read_input_file(file_name):
+    with open(file_name, "r") as input_file:
+        input_list = []
+        for line in input_file.readlines():
+            line.strip()
+            input_list.append(int(line))
+        input_list[0] = 2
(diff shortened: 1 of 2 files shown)


exercise_7.py
```diff
@@ -0,0 +1,71 @@
+from matplotlib import pyplot as plt
+import pathlib
+import numpy as np
+import Int_Computeromputer
```
joh-lin, line 4: You probably wanted to write
```python
import Int_Computer
```

exercise_7.py
```diff
@@ -0,0 +1,71 @@
...
+        return -1
+    elif ball_pos > paddle_pos:
+        return 1
+    else:
+        return 0
+
+
+def main():
+    input_list = read_input_file("breakout_commands.txt")
+    computer_test = Int_Computeromputer.IntComputer(control_paddle, insert_list)
```
joh-lin, line 65: You probably wanted to write
```python
computer_test = Int_Computer.IntComputer(control_paddle, insert_list)
```

exercise_7.py
```diff
@@ -0,0 +1,71 @@
...
+        elif x == -1:
+            global score
+            score = entity
+            continue
+        matrix[y, x] = entity
+
+
+def control_paddle():
+    global first_frame
+    insert_matrix(output)
```
joh-lin, line 46: Before `plt.imshow()` you can call
```python
plt.cla()
```
This will prevent your program from slowing down the longer it runs.

PR #197, opened 2023-11-30
Exercise_08

CorvinAn97:


FritzRa (changes requested):
All in all fairly good choice for data, however as far as i can see the data was never analysed, so the task was not completed fully.
A couple sentences Summing up the data would be great.

FritzRa (approved):
Looks good now

PR #198, opened 2023-11-30
Robert walz woche08

LeMixer:


TomKnoden (changes requested):
Surely a good solution. But you should add the given task in form of a comment above the code for context.

LeMixer:
> given task in form of a comment above the code for context.

I see the task as bloat. It makes the file cluttered and harder to review. I did not include the task in the last weeks as well for the same reason. The file is clearly named and the exercise-file is accessible for everyone curious about the task.

PR #201, opened 2023-11-30
Bo Wichtendahl, Laura van Uffelt 08

thecodingcube77:
This was done in pair programming with Laura. The same changes can also be seen in PR #200 from branch [cf32ad8](https://github.com/dhbw-ma-ppp/ppp-2023/commit/cf32ad8b2f469dcbf67c1dd75b5c6014a57b1388).


UtkuDHBW (approved):
Overall looks like a nice solution to me! Also nice use of the violin plot.

PR #205, opened 2023-11-30
exercise 8

christianrietzsch:


Exercises/ChristianRietzsch/exercise_08.ipynb
GeFritz, line 1: Nice choice of criteria. You have clear and decisive results.

You left some commented code at the end. 
```#data['Name_Length'] = data['Name'].apply(lambda x: len(x))
#data = data.sort_values('Name_Length')
#name_length = data.groupby('Name_Length')['Survived']
#name_length.count().plot()
#name_length.sum('Survived').plot(title="E")
```

This should be removed.

This might not be necessary, but you could add a bit more information about each graph and what you derive from it.

christianrietzsch:
> Nice choice of criteria and graphs. Why did you also convert to html? Seems redundant. Only necessary change is removing the commented code.

i used html because you need certain packages to convert to a pdf
and the mirror i use is broken

PR #206, opened 2023-11-30
Celina 08

CelinaRiegel:


paulseidel08 (changes requested):
You chose good and relevant statistics, but I don't know exactly, what you wanted to analyze with the first two tables. Please write some texts for the statistics containing the following information: 
1) What does the Statistic show?
2) Why is the Statistic relevant?
3) Analyze the data.
That's part of the exercise too.

PR #208, opened 2023-12-05
Gerrit Fritz Aufgabe 9: 0.60

GeFritz:


fabischw:
Requirements should be stated in requirements.txt . You should also consider using a train-test split to get validation data to validate your model BEFORE testing it on the actual test data.

GeFritz:
> Requirements should be stated in requirements.txt . You should also consider using a train-test split to get validation data to validate your model BEFORE testing it on the actual test data.

Will change reqs, grid cross validation was tested beforehand, to find best hyperparameters. Then trained on the complete trainingset.

FritzRa (approved):
Code looks good to me, also nice MCC value.

PR #210, opened 2023-12-05
Exercises 09 - Raphael and Paul (0.56)

paulseidel08:
We did Pairprogramming. So every changes will be committed in the PR of Raphael.

### DIFF
--- a/Exercises/PaulSeidel/P.Seidel_exercises_09.ipynb
+++ b/Exercises/PaulSeidel/P.Seidel_exercises_09.ipynb
@@ -1,0 +1,847 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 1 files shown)


GeFritz (commented):
Nice model. You chose a simple but efficient approach to tackle the titanic survivability predictions. 
You could try to optimise it further by for example splitting the gender into different features with boolean values. Or you could try to remove noise in the age dataset by dividing into age classes.

GeFritz (approved):
Nice model. You chose a simple but efficient approach to tackle the titanic survivability predictions.
You could try to optimise it further by for example splitting the gender into different features with boolean values. Or you could try to remove noise in the age dataset by dividing into age classes.

PR #212, opened 2023-12-06
finished_exercise_09 (0.5482001098230149)

TrueEsel:


Reisii (approved):
Looks good to me

Exercises/JanMerz/exercises_09.py
```diff
@@ -0,0 +1,67 @@
...
+# If you decide to change your code after code-review, you can report new values in the comments, but
+# leave the initial MCC in the title unchanged.
+
+import pathlib
+import pandas as pd
+from sklearn.model_selection import GridSearchCV
+from sklearn.metrics import matthews_corrcoef
+from sklearn.ensemble import RandomForestClassifier
+
+root_dir = pathlib.Path().absolute().parents[1]
```
fabischw, line 26: Consider renaming the variable to something like repo_root_dir or repo_dir or project_dir as root dir usually refers to the root directory (=highest directory in the tree ) of an operating system which could lead to confusion later on.

Exercises/JanMerz/exercises_09.py
```diff
@@ -0,0 +1,67 @@
...
+with open(root_dir / "data" / "titanic_train.csv", "r") as ifile:
+    unprocessed_data = pd.read_csv(ifile)
+
+def preprocess_data(data):
+    data = data.drop(columns = data.columns[[0, 3, 6, 7, 8, 9, 10, 11]])
```
fabischw, line 31: why load the entire dataset if you're going to delete data anyways. You can use the use 'usecols' argument when reading the data to specifiy which columns you want to load. Check https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html for more information.

Exercises/JanMerz/exercises_09.py
```diff
@@ -0,0 +1,67 @@
...
+
+    data = data.replace("female", 0)
+    data = data.replace("male", 1)
```
fabischw, line 34: This could be done easier in one line:
``py
data['Sex'] = data['Sex'].map({'female': 0, 'male': 1})
```
Not a requested change tho, just looks better and has the added benefit of only changing the one specified column (hence not searching every column and being technically faster).

File: Exercises/JanMerz/exercises_09.py [hunk 4]
```diff
@@ -0,0 +1,67 @@
...
+print(f"MCC for Train csv: {matthews_corrcoef(y, grid.predict(x))}")
+
+with open(root_dir / "data" / "titanic_test.csv", "r") as ifile:
+    test_data = pd.read_csv(ifile)
+
+processed_test_data = preprocess_data(test_data)
+
+x = processed_test_data.drop(columns = "Survived") 
+y = processed_test_data[["Survived"]]
+print((matthews_corrcoef(y, grid.predict(x))))
```
fabischw, line 67: Consider formatting this using an f-string just like you did for the Train print statement.

fabischw (changes requested):
The variable name should be changed before merging this as it could lead to a lot of confusion if it stays unchanged.

PR #214, opened 2023-12-07
Robert Walz Woche09 MCC: 0.5835

LeMixer:
Pair programming with @fabischw 
und auch einfach kompletter Murks


DrKeks1010 (changes requested):
Please update the pull request, as it seems like the last two paragraphs of code, found in Fabian's pull request, are missing here.

Overall really great solution. Well done.

Are you sure though, that in block In[128] (Fabian's version), the test for nan is working this way? Np.nan's might also compare false to np.nan

LeMixer:
> Please update the pull request, as it seems like the last two paragraphs of code, found in Fabian's pull request, are missing here.
> 
> Overall really great solution. Well done.
> 
> Are you sure though, that in block In[128] (Fabian's version), the test for nan is working this way? Np.nan's might also compare false to np.nan

Somewhere along the shenanigans I did last week on Github a couple lines of code got lost apparently. The na check works though.

DrKeks1010 (approved):
Great Job

PR #218, opened 2023-12-07
Johannes Lingk Week 9 (MCC: 0.647 for SGDClassifier)

joh-lin:


HenrietteNeumann:
nice solution

fabischw (approved):
Beatiful solution, nice column operations. But please don't use an relative path using the current working directory as reads could fail if people don't run the file from the same cwd as you do. Highly recommend building a path like this:
```py
here = pathlib.Path(os.path.abspath(''))
exercise_dir = here.parent
```

fabischw:

PR #223, opened 2023-12-07
H.Neu_ex.9

HenrietteNeumann:
MCC: 0.4958006391302672

### DIFF
--- a/exercise_9_H.Neu.py
+++ b/exercise_9_H.Neu.py
@@ -1,0 +1,67 @@
+# This is the continuation of last weeks exercise.
+# After having analysed the Titanic Dataset, you should now prepare a machine
+# learning model to predict whether passengers will survive.
+#
+# It is entirely up to you which algorithm and feature engineering to use.
+# I do recommend using some of the algorithms available in sklearn, but if you would 
+# like to use another library that's also ok. It's a good idea to try and evaluate different algorithms,
+# and different pre-processing/cleaning/feature-generation options if you have the time.
+#
+# I have split the training data into a train- and a test-set already. These can be found as separate files 
+# in the `data`-directory. You should only use the training set throughout your entire development -- feel free 
+# to use cross-validation or split the training set into a train- and a validation set again.
+# Once you have developed a final model, you should evaluate this model on the test set I've provided,
+# and report the MCC score for the test set in the title of your PR. You should **not** evaluate the test
+# set more than once for this initial submission!
+#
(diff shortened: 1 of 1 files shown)


exercise_9_H.Neu.py
```diff
@@ -0,0 +1,67 @@
...
+# If you decide to change your code after code-review, you can report new values in the comments, but
+# leave the initial MCC in the title unchanged.
+
+import pathlib
+import pandas as pd
+from sklearn.model_selection import GridSearchCV
+from sklearn.metrics import matthews_corrcoef
+from sklearn.ensemble import RandomForestClassifier
+
+root_dir = pathlib.Path().absolute().parents[1]
```
fabischw, line 26: Consider renaming the variable to something like repo_root_dir or repo_dir or project_dir as root dir usually refers to the root directory (=highest directory in the tree ) of an operating system which could lead to confusion later on.

exercise_9_H.Neu.py
```diff
@@ -0,0 +1,67 @@
...
+with open(root_dir / "data" / "titanic_train.csv", "r") as ifile:
+    unprocessed_data = pd.read_csv(ifile)
+
+def preprocess_data(data):
+    data = data.drop(columns = data.columns[[0, 3, 6, 7, 8, 9, 10, 11]])
+
+    data = data.replace("female", 0)
+    data = data.replace("male", 1)
```
fabischw, line 34: This could be done easier in one line:
``py
data['Sex'] = data['Sex'].map({'female': 0, 'male': 1})
```
Not a requested change tho, just looks better and has the added benefit of only changing the one specified column (hence not searching every column and being technically faster).

File: exercise_9_H.Neu.py [hunk 3]
```diff
@@ -0,0 +1,67 @@
...
+print(f"MCC for Train csv: {matthews_corrcoef(y, grid.predict(x))}")
+
+with open(root_dir / "data" / "titanic_test.csv", "r") as ifile:
+    test_data = pd.read_csv(ifile)
+
+processed_test_data = preprocess_data(test_data)
+
+x = processed_test_data.drop(columns = "Survived") 
+y = processed_test_data[["Survived"]]
+print((matthews_corrcoef(y, grid.predict(x))))
```
fabischw, line 67: Consider formatting this using an f-string just like you did for the Train print statement.

fabischw:
No idea why it always highlighted 4 lines (except 33,34), only meant the last lines

PR #227, opened 2023-12-14
removed unnecessary file

LeMixer:
Someone, I don't know who added an unnecessary file to the data folder.


fabischw:
The file was meant as a blueprint for other users because I saw a lot of imports not using pathlib and instead depending on the current working directory. I do agree with you that the file is in the wrong directory tho and recommend shifting the file to my personal folder and re-naming it to ‚pathlib_example.py‘ to mark the purpose of the file and avoid any further confusion.

LeMixer:
> The file was meant as a blueprint for other users because I saw a lot of imports not using pathlib and instead depending on the current working directory. I do agree with you that the file is in the wrong directory tho and recommend shifting the file to my personal folder and re-naming it to ‚pathlib_example.py‘ to mark the purpose of the file and avoid any further confusion.

Not only is the data dir the wrong directory for examples, but I don't think your personal *exercise* folder ist the right place for *blueprints*.

fabischw:
> > The file was meant as a blueprint for other users because I saw a lot of imports not using pathlib and instead depending on the current working directory. I do agree with you that the file is in the wrong directory tho and recommend shifting the file to my personal folder and re-naming it to ‚pathlib_example.py‘ to mark the purpose of the file and avoid any further confusion.
> 
> Not only is the data dir the wrong directory for examples, but I don't think your personal _exercise_ folder ist the right place for _blueprints_.

Already told you that the file is in the wrong folder so there‘s no need to re-iterate information.
I do think my personal exercise directory makes sense here as I do not have permission by the repository owner to mess with the exercise directory itself so putting a blueprint I intend on using in CR I issue makes sense to put in my exercise folder which structure is everyone‘s persona responsibility to manage.
The Exercise folder is everyone‘s ‚own‘ space and this is part of a file import so it makes sense to put it next to the exercises.

I would therefore reinstate my point of simply moving the file into my personal folder as I accidentally put it in the wrong folder and I still intend on using the blueprint to quote it in future CRs.

fabischw (approved):
Just going to re-write the template according to my specific needs and place it in a dedicated folder.

LeMixer:
> > > The file was meant as a blueprint for other users because I saw a lot of imports not using pathlib and instead depending on the current working directory. I do agree with you that the file is in the wrong directory tho and recommend shifting the file to my personal folder and re-naming it to ‚pathlib_example.py‘ to mark the purpose of the file and avoid any further confusion.
> > 
> > Not only is the data dir the wrong directory for examples, but I don't think your personal _exercise_ folder ist the right place for _blueprints_.
> 
> 
> I do think my personal exercise directory makes sense here as I do not have permission by the repository owner to mess with the exercise directory itself so putting a blueprint I intend on using in CR I issue makes sense to put in my exercise folder which structure is everyone‘s persona responsibility to manage.
> The Exercise folder is everyone‘s ‚own‘ space and this is part of a file import so it makes sense to put it next to the exercises.
> 
> I would therefore reinstate my point of simply moving the file into my personal folder as I accidentally put it in the wrong folder and I still intend on using the blueprint to quote it in future CRs.

If you really think that the *test.py* should be part of your solution to the tasks and it influences your solution in a good way go ahead and add it to your exercise dir. 
> I still intend on using the blueprint to quote it in future CRs.
Why would you not just quote examples from the actual solutions to the exercises of last week?

